PR #3, opened 2023-04-21
change to filter traffic with port/protocol

syncpark:
petabi/review#1395
aicers/piglet#454
* Add `port/protocol` to filter traffic in Piglet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "review-web"
-version = "0.1.0"
+version = "0.2.0-alpha.1"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


msk:
Pending: petabi/review-database#24

msk:
Pending the release of review-database 0.6.0.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "review-web"
-version = "0.1.0"
+version = "0.1.0-alpha.1"
```
msk, line 3: This should be "0.2.0-alpha.2".

msk (changes requested):

Cargo.toml
syncpark: I changed the version to "0.2.0-alpha.1"

PR #30, opened 2023-05-23
Broadcast internal/allow/block networks to all Hogs

syncpark:
* Add function to broadcast internal/allow/block networks to all Hogs
* `internal networks` is the networks of customer assigned to review node
  - When a new node with review is inserted, the customer networks of the node is broadcasted.
  - When the customer of review node is changed, the networks of new customer is broadcasted.
  - When the customer networks of review node is changed, the changed networks is broadcasted.
  - `internal networks` includes all of the `Intranet`, `Extranet`, `Gateway` addresses of customer.
 
The networks are defined as following:
```rust
struct HostNetworkGroup {
    hosts: Vec<IpAddr>,
    networks: Vec<IpNet>,
    ip_ranges: Vec<RangeInclusive<IpAddr>>,
}
```

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -33,8 +33,11 @@ mod traffic_filter;
 mod triage;
 mod trusted_domain;
 
+pub use self::allow_network::get_allow_networks;
+pub use self::block_network::get_block_networks;
 pub use self::cert::ParsedCertificate;
-pub use self::node::get_node_settings;
+pub use self::customer::get_customer_networks;
+pub use self::node::{get_customer_id_of_review_host, get_node_settings};
 use async_graphql::{
     connection::{Connection, Edge, EmptyFields},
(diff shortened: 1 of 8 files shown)


src/graphql.rs
```diff
@@ -33,8 +33,11 @@ mod traffic_filter;
 mod triage;
 mod trusted_domain;
 
+pub use self::allow_network::get_allow_networks;
+pub use self::block_network::get_block_networks;
 pub use self::cert::ParsedCertificate;
-pub use self::node::get_node_settings;
+pub use self::customer::get_customer_networks;
+pub use self::node::{get_customer_id_of_review_host, get_node_settings};
```
msk, line 40: 새로 public API에 추가된 함수들도 CHANGELOG에 기록해주세요.

PR #43, opened 2023-06-05
fix broadcast internal networks, allow/block list values to hog

syncpark:
Fix broadcast internet networks, allow/block list values to serialize with `bincode::DefaultOptions::new()`


msk (changes requested):
CHANGELOG 업데이트해주세요.

msk (approved):

PR #75, opened 2023-09-08
Update review-database to 0.18.0

minshao:


src/graphql/event.rs
```diff
@@ -502,12 +508,12 @@ impl EventTotalCount {
...
             || events.iter_forward(),
             |start| {
-                let start = i128::from(start.timestamp_nanos()) << 64;
+                let start = i128::from(start.timestamp_nanos_opt().unwrap_or_default()) << 64;
                 events.iter_from(start, Direction::Forward)
             },
         );
         let last = if let Some(end) = self.end {
-            let end = i128::from(end.timestamp_nanos()) << 64;
+            let end = i128::from(end.timestamp_nanos_opt().unwrap_or_default()) << 64;
```
syncpark, line 516: For `start` timestamp or for key values in various events, I think it's ok to set `...unwarp_or_default()` value when the timestamp overflows because there's no other choice.
```rust
// for start timestamp
let start = i128::from(start.timestamp_nanos_opt().unwrap_or_default()) << 64;

// for key values in outlier event
let timestamp = time.map(|t| t.timestamp_nanos_opt().unwrap_or_default());

// for key value in triage 
key.extend_from_slice(&time.timestamp_nanos_opt().unwrap_or_default().to_be_bytes());
```

For `end` timestamp, when the timestamp value overflows, we have more choices.
1. in event.rs, fn total_count(): return Ok(0) when timestamp overflows
2. in event.rs, fn latest(): return Err
3. and all others, set `end` to 0

I think we need to synchronize these action for the timestamp overflow.
When the `end` is used for the filtering, it's better to set the `end` to `i64::Max (i128::Max)` instead of 0 or Err.

```rust
     // in event.rs, fn total_count()
     let last = if let Some(end) = self.end {
            let end = i128::from(end.timestamp_nanos_opt().unwrap_or_default()) << 64;
            if end == 0 {
                return Ok(0);
            }
            end - 1
        } else {
            i128::MAX
        };

     // in event.rs fn latest()
    let latest = if let Some(end) = end {
        let end = i128::from(end.timestamp_nanos_opt().unwrap_or_default()) << 64;
        if end == 0 {
            return Err("invalid time `end`".into());
        }
        let end = end - 1;

       // in event/group.rs, fn event_frequency_series()
        let end = filter.end.map_or(i128::MAX, |t| {
            let end = i128::from(t.timestamp_nanos_opt().unwrap_or_default()) << 64;
            if end > 0 {
                end - 1
            } else {
                0
            }
        });
```

minshao: Thanks for the feedback! I've updated the PR and used `i128::MAX` as pointed out. Note that when `end` is originally set to 0, the behavior is not modified. i.e. `return Ok(0)` or `return Err("...")` or remain 0, for `total_count()`, `latest()`, `event_frequency_series()` respectively.

syncpark (approved):

PR #80, opened 2023-10-31
 Add search filter for ranked outliers

marioCluml:
Added separate method to search ranked outliers for search.

### DIFF
--- a/src/graphql/outlier.rs
+++ b/src/graphql/outlier.rs
@@ -1,4 +1,5 @@
 use super::{always_true, model::ModelDigest, Role, RoleGuard, DEFAULT_CONNECTION_SIZE};
+use crate::graphql::{earliest_key, latest_key};
 use anyhow::anyhow;
 use async_graphql::{
     connection::{query, Connection, Edge, EmptyFields},
@@ -11,6 +12,7 @@ use data_encoding::BASE64;
 use num_traits::ToPrimitive;
 use review_database::{types::FromKeyValue, Database, Direction, IterableMap};
 use serde::Deserialize;
+use serde::Serialize;
 use std::cmp;
 
 pub const TIMESTAMP_SIZE: usize = 8;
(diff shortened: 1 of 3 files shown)


src/graphql/outlier.rs
```diff
@@ -586,3 +609,161 @@ fn latest_outlier_key(before: &str) -> Result<Vec<u8>> {
...
+            continue;
+        };
+
+        if let Some(filter) = filter {
+            if let Some(time) = &filter.time {
+                let start = time.start;
+                let end = time.end;
+                if let Some(start) = start {
+                    if let Some(end) = end {
+                        if node.timestamp < start.timestamp_nanos_opt().unwrap_or_default()
```
kimhanbeom, line 728: The comparison condition for the start of the range must be "Greater than or equal"

src/graphql/outlier.rs
```diff
@@ -586,3 +609,161 @@ fn latest_outlier_key(before: &str) -> Result<Vec<u8>> {
...
+
+            if let Some(distance) = &filter.distance {
+                let start = distance.start;
+                let end = distance.end;
+                if let Some(start) = start {
+                    if let Some(end) = end {
+                        if node.distance < start || node.distance > end {
+                            continue;
+                        }
+                    } else if (((node.distance * 10.0).round() / 10.0) - start).abs() > 0.0 {
```
kimhanbeom, line 751: Why is the comparison condition to the start filter value written this way?

src/graphql/outlier.rs
```diff
@@ -586,3 +609,161 @@ fn latest_outlier_key(before: &str) -> Result<Vec<u8>> {
...
+            continue;
+        };
+
+        if let Some(filter) = filter {
+            if let Some(time) = &filter.time {
+                let start = time.start;
+                let end = time.end;
+                if let Some(start) = start {
+                    if let Some(end) = end {
+                        if node.timestamp <= start.timestamp_nanos_opt().unwrap_or_default()
```
kimhanbeom, line 728: The range of the filter should be `start..end` (start ≤ x < end), so the `=` in the part that compares to start should be subtracted.

kimhanbeom (changes requested):

src/graphql/outlier.rs
marioCluml: For the distance displayed in aide-web it seems like it's displayed in a precision of 1. However, the node.distance has a high precision. So I inferred that the user might search with number with precision 1 and changed the node.distance to a precision of 1. Also, since there were 0.0 distance values I decided not to use Epsilon in such cases since I would want to have the nodes exactly what the user searched. Would there be a better way to do this (for example: if start is what the user searched and 0.4 and node.distance was 0.44444....)?

kimhanbeom: I understand what you mean, but the current code will 'continue' even if the distance is greater than the start(For example: distance:22.222 , start:10.0), so it needs further modification.

marioCluml: Wouldn't the code move on to the next iteration? Since the difference would be greater than 0.0, I thought that it would skip this iteration and not push it into the vector ,then move onto the next iteration when the distance is greater than 0.0 since I don't want to break the loop yet to get all the nodes after I skip this one.

src/graphql/outlier.rs
marioCluml: Done! I changed it to node.timestamp < start

src/graphql/outlier.rs
kimhanbeom: The start and end of the filter are the range of distances you want to look up, so if you have a single condition of start, I think all distances greater than start should be targeted.

kimhanbeom (approved):

Issue #83, opened 2023-11-07
Corruption of `id` in `ranked outlier` graphql requests.


kimhanbeom:
When requesting graphql query `ranked outlier`, the `id` retrieved from rocksdb and the actual `id` returned by graphql do not match.
This seems to be a problem when a value of type `i64` corresponding to a timestamp is returned as a graphql `Int`.
To fix this, we need to change that return type to a `StringNumber<i64>`.

related issue: #37

PR #86, opened 2023-11-13
Modify `ranked outliers` distance search condition & Release 0.15.0

kimhanbeom:
- Change the distance search criteria to support a single search.
- Change the type of id in the `Rankedoutlier` field from `StringNumber` to `ID`.
- Change the part about `RankedOutlierTotalCount` to count the total count differently.
- Bump version to 0.15.0

Closes: #83, #87, #88

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 name = "review-web"
-version = "0.15.0-alpha.1"
+version = "0.15.0"
 edition = "2021"
 
 [dependencies]
 anyhow = "1"
-async-graphql = { version = "6", features = ["chrono", "string_number"] }
+async-graphql = { version = "6", features = ["chrono"] }
 async-graphql-axum = "6"
 async-trait = "0.1"
 axum = { version = "0.6", features = ["headers", "macros"] }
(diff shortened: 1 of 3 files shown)


src/graphql/outlier.rs
```diff
@@ -841,7 +850,7 @@ where
                         if node.distance < start || node.distance > end {
                             continue;
                         }
-                    } else if node.distance < start {
+                    } else if (((node.distance * 10.0).round() / 10.0) - start).abs() > 0.0 {
```
sehkone, line 853: When a user inputs a single number instead of a range, then this should pick up events with distance of `| d - user's number | < alpha`. I suggest 0.1 as alpha.

Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "review-web"
-version = "0.15.0-alpha.1"
+version = "0.14.6"
```
sehkone, line 3: 하위 호환성이 유지 되나요?

kimhanbeom: id가 `i64`에서 `ID` 타입으로 바뀌어서 쿼리 사용에 있어서 호환이 안되네요. 
해당 부분 변경하겠습니다.

MW-Kim (approved):

PR #115, opened 2024-01-25
Fix GraphQL pagination ordering

msk:
This resolved an issue in the GraphQL API where the ordering of edges was inconsistent when using `last`/`before` pagination arguments. According to the GraphQL Cursor Connections Specification, the order of edges should remain the same whether using `first`/`after` or `last`/`before`, provided all other arguments are equal. Previously, our API returned edges in reverse order when `last`/`before` was used, which was contrary to the specification.


sehkone:
Does this mean that the results of the GraphQL API will be different, and the code in AICE-web will need to change?

msk:
> Does this mean that the results of the GraphQL API will be different, and the code in AICE-web will need to change?

Yes.

sehkone (approved):

PR #116, opened 2024-01-25
Do not assume the internal format of the database

### LINKED ISSUE #45: Refactor `review_web::graphql::account` to use `Table::iter` instead of `Table::iter_forward`
### Current behavior

The current implementation of the `review_web::graphql::account` makes use of the `Table::iter_forward` method in order to iterate over accounts. This method directly exposes the internal representation of `review-database` by yielding `Box<u8>`, leading to tight coupling between the two components.

### Desired behavior

As part of our effort to increase the modularity of our system and to ensure loose coupling, we want to switch from using `Table::iter_forward` to `Table::iter`. The `Table::iter` method yields `Account` objects which are more abstract and easier to work with, and doesn't expose the internal representation of `review-database`.

### Tasks

To address this issue, the following tasks need to be completed:

1. Identify all instances in `review_web::graphql::account` where `Table::iter_forward` is used.

2. Replace those instances with `Table::iter`.

3. Update all relevant type conversions and handling code to work with `Account` instead of `Box<u8>`.

4. Make sure to refactor any dependencies of the changed parts of the code, if necessary.

5. Test the refactored code to ensure everything still works as expected and that there are no regression issues.

### Acceptance Criteria

1. All instances of `Table::iter_forward` in `review_web::graphql::account` are replaced with `Table::iter`.

2. All tests pass after the refactoring.

3. Review-database's internal representation is no longer exposed to `review-web`.

4. `review-web` functionality remains unchanged or improved.

Please ensure to follow our best practices for refactoring, including TDD where applicable, and providing clear, meaningful comments. Thanks in advance for your contributions to this task.

msk:
Avoid using `MapIterator`, which requires the internal format of the database, and use `Table<Account>::iter` instead.

Closes #45.

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -421,6 +421,16 @@ fn latest_key(before: &str) -> Result<Vec<u8>> {
     Ok(end)
 }
 
+/// Decodes a cursor used in pagination.
+fn decode_cursor(cursor: &str) -> Option<String> {
+    String::from_utf8(BASE64.decode(cursor.as_bytes()).ok()?).ok()
+}
+
+/// Encodes a cursor used in pagination.
+fn encode_cursor(cursor: &str) -> String {
+    BASE64.encode(cursor.as_bytes())
+}
+
 fn always_true(_ordering: cmp::Ordering) -> bool {
     true
 }
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -1,7 +1,7 @@
 use super::RoleGuard;
 use crate::auth::{create_token, decode_token, insert_token, revoke_token, update_jwt_expires_in};
 use async_graphql::{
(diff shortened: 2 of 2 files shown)


msk:
Although this PR can be merged independently, I would recommend that #115 should be merged first. Otherwise, the ordering of the GraphQL query `accountList` would become different from that of other paginated responses.

msk:
Since review-database-0.24.0 no longer provides `MapIterator` for the account table, this PR has to be merged *before* updating review-database to 0.24.0.

sehkone:
There is still another one using `MapIterator` in `template.rs`. But, it is irrelevant with this since it doesn't use any review-database's type, isn't it?

sehkone (approved):

PR #126, opened 2024-02-15
Check whether new filter name exists in the database to avoid accidental overwrite

### LINKED ISSUE #125: Unexpected Behavior in `replace_filter` Function: Overwrites Existing Filter Without Error
When a new filter is inserted using `filters.insert(new.name.clone(), new)` in the `replace_filter` function in graphql/filter.rs, the function will overwrite an existing filter if the `new.name` already exists in the filter collection. Since `replace_filter` is exposed through a GraphQL API as `replaceFilter`, there's a risk that users could unintentionally or maliciously delete any filter.

### Steps to Reproduce
1. Invoke the `replaceFilter` GraphQL API with a `new` filter input whose `name` matches an existing filter in the collection.
2. Note that the existing filter is overwritten without any error being returned, leading to potential data loss.

### Expected Behavior
The function should leverage the result of the `filters.insert` operation to check if a filter with the `new.name` already exists in the collection. Specifically, if `insert` returns `Some`, indicating that an existing filter was replaced, the function should return an error stating that the filter name already exists, thus preventing the new filter from overwriting the existing one.

### Actual Behavior
The function overwrites the existing filter with the `new` filter if they share the same `name`, without any form of error handling or notification to the caller.

### Environment
- Project version: 0.18.0-alpha.1

minshao:
Resolves #125


msk:
This PR needs to be rebased after #130 is resolved.

sehkone:
@minshao , could you resolve the conflicts?

msk:
> could you resolve the conflicts?

Resolved the conflict. Please use "Squash and merge" to maintain linear history in main.

PR #167, opened 2024-03-15
explicitly reject unsupported input combo for `load_edges`

minshao:


sehkone:
@sophie-cluml 
I think it's good to share this with you.

sehkone (approved):

PR #177, opened 2024-03-19
Update node interface

minshao:
pending on #169

@sophie-cluml @sehkone, could you please take a look at the modification regarding to `Node`. I went through the discussion that I have access to, but I am not entirely confident that I grasped all the recent design changes to the `Node` struct and the update algorithm. Please let me know if further modification is needed. Thanks.


sehkone:
@sophie-cluml , do you think this should get changed according to #177?

sophie-cluml (approved):
LGTM

Issue #146, opened 2024-02-28
Make unit tests for `node_status_list` and `account_list`


marioCluml:
Just as unit tests were added for `customers` make unit tests with such cases as:

-  last:10
-  last:10 && before
-  last:10 && after
-  first:10
-  first: 10 && before
-  first: 10 && after

so that ordering can be checked.

PR #178, opened 2024-03-20
Add unit test for nodeStatusList

marioCluml:
Added unit test for `nodeStatusList`.  It seems like `accountList` already has unit test in regards to pagination ordering.

Closes: #146 ; Closes: #180 ;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 
 - `AgentManager::halt` method to shut down a host.
+- Add unit test for `nodeStatusList` to check ordering of nodes and edges.
 
 ### Changed
 
--- a/src/graphql/node/status.rs
+++ b/src/graphql/node/status.rs
@@ -164,3 +164,312 @@ async fn load(
         }));
     Ok(connection)
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::graphql::TestSchema;
+
+    #[tokio::test]
+    async fn check_node_status_list_ordering() {
+        let schema = TestSchema::new().await;
+
+        // Insert 5 nodes
+        let res = schema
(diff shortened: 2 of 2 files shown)


sehkone:
According to #167, not all combinations are allowed. Does this reflect it?

marioCluml:
> According to #167, not all combinations are allowed. Does this reflect it?

@sehkone Two of the conditions does not follow that rule for the unit test. I will adjust it!

marioCluml:
@sehkone For `nodeStatusList`, it seems the change doesn't effect this GraphQL query. So, it causes an error if I assert for two conditions: (first, before) and (last , after). These two should give an error but I think the queries only applied to the ones in the ChangeLog. Should I remove these conditions from `nodeStatusList` test?

Screenshot 2024-03-19 at 11 52 27 PM

Screenshot 2024-03-19 at 11 52 35 PM

sophie-cluml (approved):
As discussed over huddle between Mario and Sophie, Mario push-forced a new commit with unwanted parameter combinations removed from the patch. Examples of those removed are (first, before) and (last, after).

However, as Mario indicated https://github.com/aicers/review-web/pull/178#issuecomment-2008807005 in his comment,  #167 does not influence `nodeStatusList` API. So, `nodeStatusList(first: 10, before: "some-hash")` is still a valid GraphQL request. To make that request rejected, we need to add the unwanted-parameter-combination-check code (or param-validation code) to the `nodeStatusList`. And we'd like to tackle that in a separate issue (#180), because we not only need that for `nodeStatusList`, but also for all the other GraphQL APIs that are not covered by #167.

Since the main purpose of this PR and the issue is to add test code that checks the correctness of ordering, and the current state of PR patch does not hamper the modification that is expected to be brough for #167, I am up to merging this patch to the main branch.  @sehkone, I'd like to know your thoughts on this.

Issue #144, opened 2024-02-26
Refactor `AgentManager` Trait to Enhance Encapsulation and Remove Dependency on `oinq`


msk:
* Pending the 0.20.0 release of review-web.

The current design of the `AgentManager` trait in the `review-web` crate includes the `send_and_recv` function, which allows encoding and sending arbitrary messages to agents. This design exposes the internal communication protocol with agents, which should be abstracted away within `review`, violating the encapsulation principle. To address it, this issue proposes a refactor of `AgentManager` by introducing more specific functions for each required communication pattern with agents and removing the `send_and_recv` method, and thus the dependency on `oinq`.

## Proposal:

Our refactor plan involves three main steps:

1. **Addition of Specific Functions to `AgentManager` in `review-web`:**
   - [x] Extend `AgentManager` with additional functions that encapsulate specific message types and interactions needed by `review-web` to communicate with agents. Initially, these functions will return an `Error`, serving as placeholders.
     - #148
     - #165
     - #175
   - [x] #161

2. **Implementation of the New Functions in `review`:**
   - [x] Following the `review-web` update, implement the newly added functions in `review`, adhering to the actual communication protocols with agents.

3. **Refactoring `review-web` to Utilize the New Functions:**
   - [X] With the updated `review` in place, refactor `review-web` to replace all instances of `send_and_recv` with calls to the new, more specific functions. Then, remove the `send_and_recv` method from `AgentManager`, finalizing the encapsulation enhancement.
     - #191
     - #193

## Rationale:

This refactor aims to:
- Enhance encapsulation and abstraction, ensuring `review-web` does not need to know about the specific protocol details for communicating with agents.
- Improve code maintainability and clarity by defining clear interfaces for agent communication.

PR #202, opened 2024-04-10
Remove unused function from `AgentManager`

msk:
* Pending: #201

`broadcast_to_crusher` is no longer used and shouldn't be used in the future, since it assumes review-web knows the protocol between REview and agents. See #144 for background.


msk:
Rebased on main (4c28629a) to reduce diff.

sophie-cluml (approved):
LGTM

PR #210, opened 2024-04-19
Add resetAdminPassword Mutation with Local IP Restriction

### LINKED ISSUE #176: Add `resetAdminPassword` Mutation with Local IP Restriction
We need to introduce a new GraphQL mutation, `resetAdminPassword`, which allows an admin to set a new password in the case of a forgotten one. For enhanced security, this mutation should only be accessible from a local IP address. This task will involve modifications to the routing to pass the client's remote address to our GraphQL handler, updating our `Role` enum to include a local role, and ensuring our new mutation is guarded to allow execution from local connections only.

#### Implementation Details

1. **Axum Router Configuration**: Utilize [`Router::into_make_service_with_connect_info`](https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info) method to modify the router. This change allows Axum to pass the client's remote address (`SocketAddr`) to our handlers. An example of how to use this method can be found in the Axum documentation linked above.

2. **Handler Signature Update**: Update the `graphql_handler` function to accept `ConnectInfo<SocketAddr>` as one of its extensions. This addition will enable us to access the client's remote address within our GraphQL handler and make decisions based on it.

3. **Role Enum Expansion**: Add a new variant, `Role::Local`, to our `enum Role`. This role will represent requests originating from a local IP address and will bypass the usual authentication token requirement.

4. **Local IP Address Check**: Inside the `graphql_handler`, implement logic to check if the `addr` (extracted from `ConnectInfo<SocketAddr>`) is a local IP address. If it is, assign `Role::Local` as the user's role for the duration of the request. This check effectively ignores the authentication token for local requests.

5. **Mutation Access Control**: For the `resetAdminPassword` mutation function, ensure that access is restricted to local connections by applying the `#[graphql(guard = "RoleGuard::new(Role::Local)")]` attribute. This guard ensures that the mutation can only be executed when the role is set to `Role::Local`, i.e., the request originates from a local IP address.

Please ensure thorough testing is conducted to validate the local IP address restriction works as expected, and that no regression issues are introduced with these changes.

minshao:
Resolves #176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   combinations of first, last, before, and after and apply it to GraphQL
   queries with pagination.
 - Added `LockyRansomware` detection event.
+- Added GraphQL query `resetAdminPassword` to allow resetting the password for an
+  existing user categorized as `SystemAdministrator` for administrators utilizing
+  the local network. This feature enhances the security and accessibility of user
+  accounts, providing administrators with a streamlined method for password management.
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ anyhow = "1"
 async-graphql = { version = "6", features = ["chrono"] }
 async-graphql-axum = "6"
 async-trait = "0.1"
-axum = { version = "0.6", features = ["headers", "macros"] }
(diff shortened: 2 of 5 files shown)


sehkone:
This looks good to me, except that it doesn't cover `codecov`. It would be helpful if you could fix the failure.

sophie-cluml (approved):
LGTM

Issue #37, opened 2023-05-31
`Int` Type Precision Issue in review-web's GraphQL API


msk:
There has been an issue identified concerning the data precision within review-web's GraphQL API related to integer types. At the moment, our GraphQL API employs the `Int` type, a 32-bit signed integer. However, our internal system uses Rust types such as `u32`, `isize`, `usize`, `i64`, and `u64`, which are larger than `i32` and thus, may not be properly converted to `Int`. This mismatch could lead to unintended data truncation or overflow errors when handling large values.

We propose conducting a comprehensive review of all instances in review-web's GraphQL API where the `Int` type is used. This review will evaluate if the `i32` type provides adequate precision for our current needs. If it does, we could refactor the API implementations to use `i32` instead of larger integer types, aligning the internal precision with that of the GraphQL `Int` type.

However, if we find that `i32` does not provide sufficient precision and larger integer types are necessary, it would be advisable to transition to using `StringNumber` provided by async-graphql. This would ensure the necessary precision while preserving compatibility with the GraphQL specification. Please note that this shift would require clients to handle the conversion of `StringNumber` to their respective integer types.

This issue is high priority as it can significantly affect the accuracy and reliability of the data provided to our users. It also presents an opportunity for system optimization and efficient memory usage.

The proposed steps for resolution are:

1. Conduct a comprehensive review of all instances where `Int` is used in review-web's GraphQL API.
2. Determine if the precision provided by `i32` meets the requirements of each use case.
3. If `i32` is adequate, refactor the relevant code to use `i32` rather than larger integer types.
4. If `i32` does not provide the necessary precision, transition to using the `StringNumber` type provided by async-graphql.
5. Perform extensive testing to ensure no precision-related issues emerge after the transition.

henry0715-dev:
- [ ] #259
- [ ] #260
- [ ] #261 
- [ ] #262 
- [ ] #263

henry0715-dev:
aice-web 대응 작업
https://github.com/aicers/aice-web/issues/578

sehkone:
> aice-web 대응 작업 [aicers/aice-web#578](https://github.com/aicers/aice-web/issues/578)

@henry0715-dev 

Since review-web is public and AICE-web is private, this link should be removed according to our rules. Moreover, this issue was specifically aimed at modifying review-web, and since that has been completed, the issue should have been closed.

On the other hand, in AICE-web, we can add this link so that people can recognize it as the related issue.

[...]

PR #217, opened 2024-04-30
Update `outlier` interface

minshao:
The PR adjusts `review-web` to the updated `outlier` map interface from `review-database` where `Map` is no longer exposed and the key is stored differently. To fit in, the following modifications have been made.
* `load_nodes`, `load_nodes_with_filter`, `load_with_filter`, `iter_to_nodes_with_filter` are refactored for clarity.
* `validate_and_process_pagination_params` as a validation process for `load_edges` and `load_edges` is excluded from individual query (before the load function call) to avoid duplicated decoding error.
* Switches cursor encoding format from `encode(before):encode(after)` to `encode(serialize(&(before,after)))` for security for `outlier::load`.
* Refactors `outlier::load_ranked_outliers_with_filter`, `outlier::load_with_all_outlier` and functions related for simplification

### DIFF
--- a/src/graphql/outlier.rs
+++ b/src/graphql/outlier.rs
@@ -1,27 +1,22 @@
-use super::{always_true, model::ModelDigest, Role, RoleGuard, DEFAULT_CONNECTION_SIZE};
+use super::{encode_cursor, model::ModelDigest, Role, RoleGuard};
 use crate::graphql::validate_and_process_pagination_params;
-use crate::graphql::{earliest_key, latest_key};
-use anyhow::anyhow;
 use async_graphql::{
     connection::{query, Connection, Edge, EmptyFields},
     types::ID,
-    ComplexObject, Context, InputObject, Object, ObjectType, OutputType, Result, SimpleObject,
-    Subscription,
+    ComplexObject, Context, InputObject, Object, Result, SimpleObject, Subscription,
 };
 use bincode::Options;
 use chrono::{offset::LocalResult, DateTime, NaiveDateTime, TimeZone, Utc};
-use data_encoding::BASE64;
 use futures::channel::mpsc::{unbounded, UnboundedSender};
 use futures_util::stream::Stream;
 use num_traits::ToPrimitive;
-use review_database::{types::FromKeyValue, Database, Direction, IterableMap, Store};
+use review_database::{Database, Direction, Store, UniqueKey};
 use serde::Deserialize;
 use serde::Serialize;
-use std::{cmp, collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc};
 use tokio::{sync::RwLock, time};
 use tracing::error;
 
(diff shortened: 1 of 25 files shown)


msk:
Requires #219 to be merged first to pass CI.

src/graphql/outlier.rs
```diff
@@ -462,36 +408,16 @@ struct OutlierTotalCount {
 #[Object]
 impl OutlierTotalCount {
     /// The total number of edges.
-    async fn total_count(&self, ctx: &Context<'_>) -> Result<i64> {
-        let prefix = bincode::DefaultOptions::new().serialize(&self.model_id)?;
+    async fn total_count(&self, ctx: &Context<'_>) -> Result<usize> {
```
sehkone, line 411: I was wondering if we could use `usize` since GraphQL doesn't support a 4-byte unsigned integer type. For the total number of counts, we use `usize` in some places but `i64` elsewhere. @minshao, could you enlighten me on this?

msk: The return type in this instance should be refactored to either `i32` or `StringNumber` to conform to the GraphQL type system, which utilizes `Int` as a 32-bit signed integer. This adjustment is necessary to address the data precision issues outlined in issue #37. This recommendation is not specifically for this PR alone but is a broader action item we need to apply throughout the codebase to ensure consistency and prevent potential data truncation or overflow errors.

msk (approved):

src/graphql/statistics.rs
```diff
@@ -45,8 +44,6 @@ impl StatisticsQuery {
         last: Option<i32>,
     ) -> Result<Connection<String, Round, TotalCountByCluster, EmptyFields, RoundByCluster>> {
         let cluster = cluster.as_str().parse()?;
-        let (after, before, first, last) =
-            validate_and_process_pagination_params(after, before, first, last)?;
```
sophie-cluml, line 49: It seems like the validation for the params is not handled afterwards for this endpoint.

minshao: Later in the function it called `load_rounds_by_cluster` which includes:
``` Rust
  let after = slicing::decode_cursor(after)?.map(|(_, t)| i64_to_naive_date_time(t));
  let before = slicing::decode_cursor(before)?.map(|(_, t)| i64_to_naive_date_time(t));
```
The removal here is actually to fix the double decoding issue in the original code, since `validate_and_process_pagination_params` also decodes `after` and `before`. 
In addition, if `validate_and_process_pagination_params` is preferred for PostgresQL related parameter validation, then please create an issue and I'll attend to it in a separate PR.

sophie-cluml: Thank you for clarifying it. I think input parameter validation is something to be run before the logic proceeds to do DB related `load_*` functions, so let me create another issue for this.

sophie-cluml (approved):
LGTM

sehkone:
@minshao could you resolve the conflict?

PR #219, opened 2024-05-02
Use `clone_from` instead of `clone`

minshao:


msk:
This PR fixes clippy warnings with Rust 1.78.

sehkone (approved):

PR #235, opened 2024-06-17
Add specific type definitions for `StringNumber` in `NodeStatus`

### LINKED ISSUE #231: Must specify number types in documentation comment when using `StringNumber` in GraphQL API responses
### Background

Recently, https://github.com/aicers/review-web/issues/208 changed the `totalMemory`, `usedMemory`, `totalDiskSpace`, `usedDiskSpace`, `ping` to `StringNumber`.

The GraphQL schema in SDL format we get looks like below. This means the information of `StringNumber`'s number type is lost. This happened even though we specified the types of `totalMemory`, `usedMemory`, `totalDiskSpace`, `usedDiskSpace`, `ping` as `StringNumber<u64>`, `StringNumber<u64>`, `StringNumber<u64>`, `StringNumber<u64>`, `StringNumber<i64>`, respectively.

```graphql
type NodeStatus {
  # The hostname of the node.
  name: String!

  # The average CPU usage in percent.
  cpuUsage: Float

  # Whether review is online or not.
  review: Boolean

  # Whether piglet is online or not.
  piglet: Boolean

  # actual piglet configuration
  pigletConfig: PigletConfig

  # Whether reconverge is online or not.
  reconverge: Boolean

  # actual reconverge configuration
  reconvergeConfig: ReconvergeConfig

  # Whether hog is online or not.
  hog: Boolean

  # actual hog configuration
  hogConfig: HogConfig
  id: ID!
  totalMemory: StringNumber
  usedMemory: StringNumber
  totalDiskSpace: StringNumber
  usedDiskSpace: StringNumber
  ping: StringNumber
}

scalar StringNumber
```

This calls for the need to correctly specify the number types of the fields that are represented as `StringNumber`, so that the client side of this GraphQL API can appropriately know into which number type it can parse the fields.

### Tasks
1. Specify the number types of above 5 fields in documentation comment.
2. Be aware that we should ensure we specify the number types when using `StringNumber`.

div-seungha:
resolve #231 

In this commit, the schema of the graphql playground is as follows.

image

### DIFF
--- a/src/graphql/node.rs
+++ b/src/graphql/node.rs
@@ -462,23 +462,23 @@ impl NodeStatus {
     async fn id(&self) -> ID {
         ID(self.id.to_string())
     }
-
+    /// The RAM size in bytes within the range representable by a `u64`
     async fn total_memory(&self) -> Option<StringNumber<u64>> {
         self.total_memory.map(StringNumber)
     }
-
+    /// The amount of used RAM in bytes within the range representable by a `u64`
     async fn used_memory(&self) -> Option<StringNumber<u64>> {
         self.used_memory.map(StringNumber)
     }
-
+    /// The total disk space in bytes within the range representable by a `u64`
     async fn total_disk_space(&self) -> Option<StringNumber<u64>> {
         self.total_disk_space.map(StringNumber)
     }
-
+    /// The total disk space in bytes that is currently used within the range representable by a `u64`
     async fn used_disk_space(&self) -> Option<StringNumber<u64>> {
         self.used_disk_space.map(StringNumber)
     }
-
+    /// The round-trip time in microseconds to a host, within the range representable by an `i64`
     async fn ping(&self) -> Option<StringNumber<i64>> {
         self.ping.map(StringNumber)
     }


src/graphql/node.rs
```diff
@@ -461,23 +462,23 @@ impl NodeStatus {
...
         self.total_disk_space.map(StringNumber)
     }
-
+    /// The total disk space in bytes that is currently used. The type must be u64.
     async fn used_disk_space(&self) -> Option<StringNumber<u64>> {
         self.used_disk_space.map(StringNumber)
     }
-
+    /// The ping value for a specific node. The type must be u64.
     async fn ping(&self) -> Option<StringNumber<i64>> {
```
sehkone, line 482: The types specified in the comment and the return value are not the same.

div-seungha: I missed it. 🥲  I replaced `u64` to `i64`.

sophie-cluml (approved):
lgtm

sehkone:
@div-seungha 
According to this [rule](https://www.notion.so/cluml/af1d5c95d1064bfc8eb8474b81b1ca8b?pvs=4#280beba87c9447869bcb87b7f647853e), I think it would be better to remove the recent commit.

div-seungha:
> @div-seungha According to this [rule](https://www.notion.so/cluml/af1d5c95d1064bfc8eb8474b81b1ca8b?pvs=4#280beba87c9447869bcb87b7f647853e), I think it would be better to remove the recent commit.

I made a new commit and the prior commits were removed by force-push.

src/graphql/node.rs
```diff
@@ -461,23 +462,23 @@ impl NodeStatus {
     async fn id(&self) -> ID {
         ID(self.id.to_string())
     }
-
+    /// The RAM size in bytes. The type must be u64.
```
sehkone, line 465: @sophie-cluml @div-seungha 
Users of GraphQL APIs will read this comment via the playground. It means this is supposed to explain what kind of value(s) the API returns. Therefore, I suggest:

```
/// The RAM size in bytes within the range representable by a `u64`
```

div-seungha: I wasn't sure if the comment contained a valid or right content. I proceeded with the modification as you suggested and I think your suggestion is much useful for users to understand the meaning clearly.

src/graphql/node.rs
```diff
@@ -461,23 +462,23 @@ impl NodeStatus {
...
     async fn total_disk_space(&self) -> Option<StringNumber<u64>> {
         self.total_disk_space.map(StringNumber)
     }
-
+    /// The total disk space in bytes that is currently used within the range representable by a `u64`.
     async fn used_disk_space(&self) -> Option<StringNumber<u64>> {
         self.used_disk_space.map(StringNumber)
     }
-
+    /// The ping value for a specific node within the range representable by a `i64`.
```
sehkone, line 481: You can find out useful information what the ping value exactly means in the code such as the following in the changelog:

```
 - `AgentManager::ping` method to measure the latency between the agent manager
  and a host.
```
Another one is written in the `ping` method as follows:
```
    /// Sends a ping message to the given host and waits for a response. Returns
    /// the round-trip time in microseconds.
    async fn ping(&self, _hostname: &str) -> Result<i64, anyhow::Error>;
```
From these, I suggest like:
```
/// The round-trip time in microseconds to a host, within the range representable by an `i64`.
```

div-seungha:
The close of this PR was my mistake. I reopened the PR. Sorry.

sophie-cluml:
@div-seungha it seems like your commit is reverted back to initial PR state. Could you check? And please rebase your PR on `main` branch.

src/graphql/node.rs
sophie-cluml: I am obviously not Sehkone, but if I may chip in, it seems fine to me to make change only for "ping". My understanding of the reasons behind Kone's request is to:
- indicate the unit as microsecond
- be kind to explain ping as round-trip time, because ping might be a bit unfriendly term for client side engineers

Documentation comments on other fields seem to be meeting these criteria already, so I think just modifying the documentation comment for "ping" is enough.

div-seungha:
> @div-seungha it seems like your commit is reverted back to initial PR state. Could you check? And please rebase your PR on `main` branch.

Sorry, there were some mistakes. I checked my commit and restored missing parts by force-push.

sophie-cluml (approved):
LGTM

src/graphql/node.rs
```diff
@@ -462,23 +462,23 @@ impl NodeStatus {
     async fn id(&self) -> ID {
         ID(self.id.to_string())
     }
-
+    /// The RAM size in bytes. The type must be u64.
     async fn total_memory(&self) -> Option<StringNumber<u64>> {
         self.total_memory.map(StringNumber)
     }
-
+    /// The amount of used RAM in bytes. The type must be u64.
```
sehkone, line 469: This is the same as what you wrote in the first place. I think it is a minor mistake, but these are the kinds of errors that waste our time.

div-seungha: I am really sorry. There were some critical mistakes when I was controlling rebase and reset commits through git command. I should have checked the diff more meticulously.

sehkone:
@div-seungha 
We usually insert an empty line between methods, but since it's not mandatory, I am merging this into the main branch. I would like you to do so in the future.

Issue #250, opened 2024-07-24
Modify related APIs due to addition of `language` field in `Account`


div-seungha:
This issue was created to support changes caused by the following issue: https://github.com/petabi/review-database/issues/308

The `language` field will be newly added to the `Account` struct in the review-database to represent the language selected by the user in the user interface. Therefore, the relevant APIs need to be modified in this repository as well.

PR #258, opened 2024-08-12
Modify related APIs due to addition of language field in Account

div-seungha:
Close: https://github.com/aicers/review-web/issues/250

Among the list of tasks I needed to perform was the following items.
- Modify account, insert_account, modify_account, and other related APIs in review-web as needed to support the field.
- Add two new APIs in review-web: one for retrieving the user's language setting, and another for updating it.

However, after looking at the existing code, I found that the original purpose could be achieved just by modifying the existing API rather than adding new APIs, so I worked on adding language-related codes to the existing function without adding separate new APIs. But, I think there could be some misunderstanding about these task so before I open the PR to merge, I have just opened the draft PR and I ask you to review the code I worked on. @sehkone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Refactor `AgentManager::ping` to return `Duration` instead of `i64`. This
   refactor improves the flexibility and accuracy of the `ping` method, making it
   more robust and aligned with Rust's time handling conventions.
+- Added a `language` field to the `Account`. Consequently, the `account` and
+  `accountList` API responses now include this field. The `insertAccount` and
+  `updateAccount` GraphQL API endpoints are also updated to support the field.
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ num-traits = "0.2"
 reqwest = { version = "0.12", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
-review-database = { git = "https://github.com/petabi/review-database.git", tag = "0.29.1" }
+review-database = { git = "https://github.com/petabi/review-database.git", rev = "66ccdd49" }
 review-protocol = { git = "https://github.com/petabi/review-protocol.git", rev = "4a6ea44" }
 roxy = { git = "https://github.com/aicers/roxy.git", tag = "0.2.1" }
 rustls = "0.23" #should be the same version as what reqwest depends on
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -194,6 +194,7 @@ impl AccountMutation {
                     &None,
                     &None,
                     &None,
+                    &None,
                 )?;
                 return Ok(username);
             }
@@ -235,6 +236,7 @@ impl AccountMutation {
         role: Option<UpdateRole>,
         name: Option<UpdateName>,
         department: Option<UpdateDepartment>,
+        language: Option<UpdateLanguage>,
         allow_access_from: Option<UpdateAllowAccessFrom>,
         max_parallel_sessions: Option<UpdateMaxParallelSessions>,
     ) -> Result<String> {
@@ -242,6 +244,7 @@ impl AccountMutation {
             && role.is_none()
             && name.is_none()
             && department.is_none()
+            && language.is_none()
             && allow_access_from.is_none()
             && max_parallel_sessions.is_none()
         {
@@ -251,6 +254,7 @@ impl AccountMutation {
         let role = role.map(|r| (database::Role::from(r.old), database::Role::from(r.new)));
         let name = name.map(|n| (n.old, n.new));
         let dept = department.map(|d| (d.old, d.new));
+        let language = language.map(|d| (d.old, d.new));
         let allow_access_from = if let Some(ip_addrs) = allow_access_from {
             let old = if let Some(old) = ip_addrs.old {
                 Some(strings_to_ip_addrs(&old)?)
@@ -276,6 +280,7 @@ impl AccountMutation {
             role,
             &name,
             &dept,
+            &language,
             &allow_access_from,
             &max_parallel_sessions,
         )?;
@@ -467,6 +472,10 @@ impl Account {
         &self.inner.department
     }
 
+    async fn language(&self) -> Option<String> {
+        self.inner.language.clone()
+    }
+
     async fn creation_time(&self) -> DateTime<Utc> {
         self.inner.creation_time()
     }
@@ -536,6 +545,12 @@ struct UpdateName {
 struct UpdateDepartment {
     old: String,
     new: String,
+}
+
+#[derive(InputObject)]
+struct UpdateLanguage {
+    old: Option<String>,
+    new: Option<String>,
 }
 
 /// The old and new values of `allowAccessFrom` to update.
@@ -1121,7 +1136,8 @@ mod tests {
                         password: "Ahh9booH",
                         role: "SECURITY_ADMINISTRATOR",
                         name: "John Doe",
-                        department: "Security"
+                        department: "Security",
+                        language: "en-US"
                     )
                 }"#,
             )
@@ -1136,7 +1152,8 @@ mod tests {
                         password: "Ahh9booH",
                         role: "SYSTEM_ADMINISTRATOR",
                         name: "John Doe",
-                        department: "Admin"
+                        department: "Admin",
+                        language: "en-US"
                     )
                 }"#,
             )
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ num-traits = "0.2"
 reqwest = { version = "0.12", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
-review-database = { git = "https://github.com/petabi/review-database.git", tag = "0.29.1" }
+review-database = { git = "https://github.com/petabi/review-database.git", rev = "c3343e3" }
```
sehkone, line 29: This commit includes changes that are unrelated to the current PR. While it can be beneficial to use the latest commit if you’re familiar with its contents, it’s generally better to use the closest relevant commit that directly supports the PR. However, when releasing a new version, we rely on the latest releases from other modules.

div-seungha: I understood what you said. I have just replaced it with a commit hash that is directly related to this PR and force-pushed it.

sophie-cluml (commented):
This draft looks good to me in regard to modifying existing APIs. 

In regard to your this question,

> However, after looking at the existing code, I found that the original purpose could be achieved just by modifying the existing API rather than adding new APIs, so I worked on adding language-related codes to the existing function without adding separate new APIs. But, I think there could be some misunderstanding about these task so before I open the PR to merge, I have just opened the draft PR and I ask you to review the code I worked on

I think we need separate APIs because current `insert_account` and `update_account` requires the user to be `SystemAdministrator` or `SecurityAdministrator`. I think we want to allow users of any role type to retrieve and modify their language settings.

sehkone:
> This draft looks good to me in regard to modifying existing APIs.
> 
> In regard to your this question,
> 
> > However, after looking at the existing code, I found that the original purpose could be achieved just by modifying the existing API rather than adding new APIs, so I worked on adding language-related codes to the existing function without adding separate new APIs. But, I think there could be some misunderstanding about these task so before I open the PR to merge, I have just opened the draft PR and I ask you to review the code I worked on
> 
> I think we need separate APIs because current `insert_account` and `update_account` requires the user to be `SystemAdministrator` or `SecurityAdministrator`. I think we want to allow users of any role type to retrieve and modify their language settings.

I already specified the additional APIs in the patio.

div-seungha:
> This draft looks good to me in regard to modifying existing APIs.
> 
> In regard to your this question,
> 
> > However, after looking at the existing code, I found that the original purpose could be achieved just by modifying the existing API rather than adding new APIs, so I worked on adding language-related codes to the existing function without adding separate new APIs. But, I think there could be some misunderstanding about these task so before I open the PR to merge, I have just opened the draft PR and I ask you to review the code I worked on
> 
> I think we need separate APIs because current `insert_account` and `update_account` requires the user to be `SystemAdministrator` or `SecurityAdministrator`. I think we want to allow users of any role type to retrieve and modify their language settings.

I've got your point. I will create new APIs so that anyone can modify the language settings regardless of the role. @sophie-cluml

div-seungha:
I added a function to update the user's language selection. Originally, the task description I was assigned was supposed to add two api for each of 'retrieve' and 'update', but I thought that just one function to change the language setting was enough, so I added just one api. (but I may have misunderstood the meaning of the word 'retrieve'. I understood the word in the meaning of 'recovery.' If so, please let me know with comments here.)

I think I need to write the test code on the function I modified and added, but I'm not very good at how to write the test code in this repository yet, especially with using tokio. If you have any advice, I'll refer to it and add the test code. @sehkone @sophie-cluml

sehkone:
@sophie-cluml @div-seungha,

I recall that this PR was originally intended to address only issue #250. However, it now also includes changes for issue #268. I believe this is not optimal. It would be more appropriate to manage the new issue in a separate PR.

PRs should be processed expeditiously. Continuing to add new features and modifications will necessitate additional discussion and feedback, which can significantly delay resolution and lead to negative consequences.

Furthermore, the initial text of this PR got modified to include this new issue. When modifying a text, any changes should be clearly visible to others. In this context, that is not easy. Therefore, if you need to modify text written earlier, it would be better to add a new comment instead of editing the existing text directly.

In conclusion, I recommend removing the additions related to issue #268 and addressing them in a separate PR.

div-seungha:
> @sophie-cluml @div-seungha,
> 
> I recall that this PR was originally intended to address only issue #250. However, it now also includes changes for issue #268. I believe this is not optimal. It would be more appropriate to manage the new issue in a separate PR.
> 
> PRs should be processed expeditiously. Continuing to add new features and modifications will necessitate additional discussion and feedback, which can significantly delay resolution and lead to negative consequences.
> 
> Furthermore, the initial text of this PR got modified to include this new issue. When modifying a text, any changes should be clearly visible to others. In this context, that is not easy. Therefore, if you need to modify text written earlier, it would be better to add a new comment instead of editing the existing text directly.
> 
> In conclusion, I recommend removing the additions related to issue #268 and addressing them in a separate PR.

I understand what you mean. I have restored the commit to the previous state and changed the current PR from Draft to ready to merge. I would appreciate it if you could check this PR as a PR to solve #250 . I will create a separate PR right away for the new API addition. @sehkone cc. @sophie-cluml

sophie-cluml:
@div-seungha test code 작성에 대해 논의하기 위해 따로 미팅 요청 드리도록 하겠습니다.

sophie-cluml:
@div-seungha CI fail 확인 부탁드립니다.

src/graphql/account.rs
```diff
@@ -1121,7 +1136,8 @@ mod tests {
                         password: "Ahh9booH",
                         role: "SECURITY_ADMINISTRATOR",
                         name: "John Doe",
-                        department: "Security"
+                        department: "Security",
+                        language: "English"
```
sophie-cluml, line 1140: Why don't we change this value to what GraphQL clients will actually use?

sophie-cluml (commented):
test code 작성 방향은 잘 된 것 같아요

src/graphql/account.rs
div-seungha: Since the UI uses the combination of ISO 639-1 and ISO 3166-1 codes such as "en-US", the value Graphql clients use will be the string like "en-US". I have just replace `"English"` and `"Korean"` with `"en-US"` and `"ko-KR"`. Did I understand your points right?

sophie-cluml: Looks awesome.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Refactor `AgentManager::ping` to return `Duration` instead of `i64`. This
   refactor improves the flexibility and accuracy of the `ping` method, making it
   more robust and aligned with Rust's time handling conventions.
+- Modified the functions that inserts and updates the account to include the language
```
sophie-cluml, line 62: Could you check this sentence's grammar with external translation tools please?

div-seungha: Sorry, I've corrected the wrong part.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Refactor `AgentManager::ping` to return `Duration` instead of `i64`. This
   refactor improves the flexibility and accuracy of the `ping` method, making it
   more robust and aligned with Rust's time handling conventions.
+- Modified the functions to insert and update accounts to include `language`
```
sophie-cluml, line 62: changelog의 예상독자는 현재 리포지토리 코드 기여자뿐만 아니라, 이 review-web을 사용하는 client developer도 해당하는 것으로 알고 있습니다. 그래서 이번 PR과 같이 API 인터페이스에 변경이 있는 경우라면 해당 API가 무엇인지 적어주면 더 좋을 것 같아요. 아래와 같이 적어준다면, insert account와 update account 하는 API 가 무엇인지 유추하지 않고 명시적으로 받아들일 수 있는 장점이 있을 것 같다는 생각이 듭니다. 그리고 `account` 와 `accountList` API 도 영향을 받았다는 것이 놓쳐지지 않도록 꼼꼼히 체크되면 좋겠습니다. 아래와 같이 (chatgpt로 생성한) 문구를 하나의 예시로서 드려보고자 합니다. 문장 마침표도 꼭 찍어주세요.

```
Added a `language` field to the `Account`. Consequently, the `account` and `accountList` API responses now include this field. The `insertAccount` and `updateAccount` GraphQL API endpoints are also updated to support the field.
```

div-seungha: 네 감사합니다. 아직 changelog에 변경사항을 기록하는 것이 익숙하지 않아서, 이번 PR에는 예시로 적어 주신 내용을 그대로 따르고 다음 PR부터는 이를 참고해서 좀 더 상세하게 작성할 수 있도록 해보겠습니다.

sophie-cluml (approved):
lgtm

Cargo.toml
```diff
@@ -26,7 +26,7 @@ num-traits = "0.2"
 reqwest = { version = "0.12", default-features = false, features = [
   "rustls-tls-native-roots",
 ] }
-review-database = { git = "https://github.com/petabi/review-database.git", tag = "0.29.1" }
+review-database = { git = "https://github.com/petabi/review-database.git", rev = "0340144" }
```
sehkone, line 29: It would be better to use `66ccdd49` because it includes the clarified README.

div-seungha: I replaced the commit hash `0340144` with `66ccdd49`.

sehkone (approved):

PR #265, opened 2024-08-16
Use `StringNumber` for integers beyond `i32` - part1

henry0715-dev:
Close #260

PR #269, opened 2024-08-20
Use `StringNumber` for integers beyond `i32` - part3

marioCluml:
Closes: #262 ;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Correct the release date of `0.20.0` to `2024-04-25`.
+- Fixed `Int` type precision error that occurs when graphql API have instances
+  that are greater than `i32` (`u32`,`u64`,`isize`,`usize`,`i64`) for
+  `NetworkThreat`,`AccountTotalCount`,`ClusterTotalCount`,`ColumnTimeSeries`,
+  `ModelIndicator`.
 
 ## [0.20.0] - 2024-04-25
 
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -6,7 +6,7 @@ use std::{
 use anyhow::anyhow;
 use async_graphql::{
     connection::{query, Connection, EmptyFields},
-    Context, Enum, InputObject, Object, Result, SimpleObject,
+    Context, Enum, InputObject, Object, Result, SimpleObject, StringNumber,
 };
 use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use review_database::{
@@ -491,8 +491,10 @@ impl Account {
             .map(|ips| ips.iter().map(ToString::to_string).collect::<Vec<String>>())
     }
 
-    async fn max_parallel_sessions(&self) -> Option<u32> {
-        self.inner.max_parallel_sessions
+    /// The max sessions that can be run in parallel in string within the
+    /// representable range of `u32`.
+    async fn max_parallel_sessions(&self) -> Option<StringNumber<u32>> {
+        self.inner.max_parallel_sessions.map(StringNumber)
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/graphql/account.rs
```diff
@@ -554,14 +555,14 @@ struct AccountTotalCount;
 
 #[Object]
 impl AccountTotalCount {
-    /// The total number of edges.
-    async fn total_count(&self, ctx: &Context<'_>) -> Result<usize> {
+    /// The total number of edges in string within the representable range of `usize`
+    async fn total_count(&self, ctx: &Context<'_>) -> Result<StringNumber<usize>> {
```
henry0715-dev, line 559: There is no need to change the part used for paging.

marioCluml: Done, removed `StringNumber` from paging for `clusterTotalCount` and `accountTotalCount`.

src/graphql/cluster.rs
```diff
@@ -294,8 +294,9 @@ struct ColumnTimeSeries<'a> {
 
 #[Object]
 impl ColumnTimeSeries<'_> {
-    async fn column_index(&self) -> usize {
-        self.inner.column_index
+    // The column index of the time series in string within the representable range of `usize`
```
henry0715-dev, line 297: `//` It would be better to change to `///`.
And it is recommended to modify the document to end with a "."

henry0715-dev:
I think it would be better to modify the Title.
Use `StringNumber` for integers beyond `i32`
example - https://github.com/aicers/review-web/pull/265

src/graphql/event/network.rs
```diff
@@ -82,8 +82,10 @@ impl NetworkThreat {
         &self.inner.service
     }
 
-    async fn last_time(&self) -> i64 {
-        self.inner.last_time
+    /// The last time the event was seen in string within the represebtable
```
henry0715-dev, line 85: "representable" appears to be a typo.

marioCluml: Done, fixed the typos

henry0715-dev (approved):
👍

sophie-cluml:
Could you rebase this please?

CHANGELOG.md
```diff
@@ -72,6 +72,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Correct the release date of `0.20.0` to `2024-04-25`.
+- Fixed `Int` type precision error that occurs when graphql API have instances
+  that are greater than `i32` (`u32`,`u64`,`isize`,`usize`,`i64`) for
+  `NetworkThreat`,`AccountTotalCount`,`ClusterTotalCount`,`ColumnTimeSeries`,
+  `ModelIndicator`.
```
sehkone, line 78: @marioCluml @henry0715-dev 

The previously merged commits documented this in the section `Changed`, but this did in the section `Fixed`. If we consider this a bug we need to put all into `Fixed`, but I think it is still Ok to keep it in the `Changed`.  Either way, we have to make it consistent.

Because the correction of README can be done right before releasing a new version, I am merging this as it is. And, I will take care of the README later.

sehkone (approved):

CHANGELOG.md
marioCluml: Yes because we were merging it sequentially I was going to rebase this branch on top of main after part 2 was merged and in doing so fix the CHANGELOG that has part 2 CHANGELOG entry. Then for part 4 I was going to do the same process as well. So that way it will be more consistent while being merged sequentially.

marioCluml: I'll be back home in about an hour or so and can fix it then since part 2 was merged earlier today. I was waiting for the part 2 to be merged before fixing this

sehkone: @marioCluml 

I suggest fixing the README in #270 instead of this. By doing so, I think we will not need to enumerate all the items and would rather mention like "all the relevant APIs" if we cover all.

PR #275, opened 2024-08-21
Refactor `AgentManager::ping` to return `Duration`

### LINKED ISSUE #274: Improve precision and flexibility of `AgentManager::ping` return value
Currently, the `AgentManager::ping` method returns the round-trip time as an `i64` representing microseconds. This approach has two limitations:

1. It forces a specific precision (microseconds) on all callers, which may not be appropriate for all use cases.
2. It potentially loses precision if the underlying measurement is more accurate than microseconds.

To address these issues, we should change the return type to `std::time::Duration`. This will:

- Preserve the full precision of the round-trip time measurement
- Allow callers to choose their desired precision when using the result
- Align with Rust standard library practices for representing time intervals

msk:
This refactor improves the flexibility and accuracy of the `ping` method, making it more robust and aligned with Rust's time handling conventions.

Resolves #274.


sophie-cluml (approved):
lgtm

PR #280, opened 2024-08-23
Update GraphQL API `applyNode`

### LINKED ISSUE #251: Update `applyNode` API behavior
### Background
https://github.com/aicers/review-web/issues/245 작업 이후, `profile`, `profile_draft`, `giganto`, `agents`, `config`, `draft` 와 같은 필드를 review-web에서도 사용할 수 있습니다.

### Tasks
- When there is an update in the configuration of a node after users click the apply button, central management server notifies the corresponding agents of the changes. We can use review-protocol crate's interface to notify the agents. At the same time, central management server updates the node's data, such as `name` with `name_draft`, `profile` with `profile_draft` ~and sets `profile_draft` to `None`~, and updates `config` with `draft`. ~For Giganto, `draft` will be set to `None`.~
- We need to log apply actions along with the `profile`, `profile_draft`, `config`, `draft` and the timestamp.

sophie-cluml:
close #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   during sign-in.
 - Added ip access control based on the `allow_access_from` field of `Account`
   during sign-in.
+- Added `AgentManager::update_config` method to notify agents to update their
+  configurations.
 
 ### Changed
 
@@ -72,6 +74,25 @@ Versioning](https://semver.org/spec/v2.0.0.html).
     `post_body`, `state`
 - Added new fields to `BlockListConn` event.
   - `orig_l2_bytes`, `resp_l2_bytes`
+- Updated the `applyNode` GraphQL API to align with the new node and agent
+  management approach.
+  - The API updates the database with draft values, notifies agents to update
+    their configurations, and logs the changes, as long as each step is needed.
+  - The `successModules` field has been removed from the API response. Instead,
+    the response now includes `gigantoDraft`, representing the draft
+    configuration of the Giganto module. If `gigantoDraft` is `None`, it means
+    either the node does not have the Giganto module or the draft for the
+    Giganto is unavailable. In the latter case, this indicates that the Giganto
+    should be disabled, resulting in the node no longer having the Giganto
+    module.
+
+### Removed
+
+- The `AgentManager::set_config` method has been removed, due to the new
+  configuration management approach. The central management server no longer
+  sends updates directly to agents. Instead, it notifies them through the
+  `update_config` method, prompting agents to request the updated configuration
+  from the management server.
 
 ### Fixed
 
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -61,13 +61,8 @@ pub trait AgentManager: Send + Sync {
     /// Reboots the node with the given hostname.
     async fn reboot(&self, _hostname: &str) -> Result<(), anyhow::Error>;
 
-    /// Sets the configuration of the given agent.
-    async fn set_config(
-        &self,
-        _hostname: &str,
-        _agent_id: &str,
-        _config: &Config,
-    ) -> Result<(), anyhow::Error>;
+    /// Notifies agents to update their configuration.
+    async fn update_config(&self, _agent_key: &str) -> Result<(), anyhow::Error>;
 
     /// Updates the traffic filter rules for the given host.
     async fn update_traffic_filter_rules(
--- a/src/graphql/node/control.rs
+++ b/src/graphql/node/control.rs
@@ -1,14 +1,13 @@
 use async_graphql::{Context, Object, Result, SimpleObject, ID};
-use review_database::{Node, NodeProfile};
+use futures::future::join_all;
+use review_database::Node;
 use tracing::{error, info};
 
 use super::{
     super::{BoxedAgentManager, Role, RoleGuard},
-    ModuleName, NodeControlMutation,
+    NodeControlMutation,
 };
 use crate::graphql::{customer::broadcast_customer_networks, get_customer_networks};
-
-const MAX_SET_CONFIG_TRY_COUNT: u32 = 3;
 
 #[Object]
 impl NodeControlMutation {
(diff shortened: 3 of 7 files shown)


src/graphql/node/control.rs
```diff
@@ -39,172 +38,231 @@ impl NodeControlMutation {
...
 }
 
-#[allow(clippy::struct_excessive_bools)]
-struct ModuleSpecificProfileUpdateIndicator {
-    hog: bool,
-    reconverge: bool,
-    piglet: bool,
+// The scope of the applying node operation for a node.
+#[allow(clippy::enum_variant_names)]
+enum NodeApplyScope<'a> {
```
sehkone, line 172: I was wondering why all the variants include Giganto.

sehkone: I mean there might be a case where the _apply_ doesn't affect Giganto.

sehkone: How about handling the three steps—name and profile, agents, and Giganto—in sequence, instead of separating the cases?

sehkone: To me, `enum` variants seem to represent mutually exclusive cases, which is not the case.

sophie-cluml: I made code changes on this point. I'd appreciate if you could check.

src/graphql/node/control.rs
```diff
@@ -39,172 +38,222 @@ impl NodeControlMutation {
...
+            let agent_manager = ctx.data::<BoxedAgentManager>()?;
+            notify_agents(
+                agent_manager,
+                &target_agents.updates,
+                &target_agents.disables,
+            )
+            .await?;
+
+            info!(
+                "[{}] Node ID {i} - Node's agents are notified to apply.\n{:?}",
```
sehkone, line 81: The term 'apply' means any updates will be applied to the node. And, each agent
will be updated with a new configuration. Therefore, I suggest the message like
"The node are notified to update its agents."

sehkone: Or, "Node's agents are notified to be updated."

src/graphql/node/control.rs
```diff
@@ -39,172 +38,222 @@ impl NodeControlMutation {
...
+                            agent.key, agent.config, agent.draft
+                        ))
+                    } else {
+                        None
+                    }
+                })
+            );
+        }
+
+        if apply_scope.db {
```
sehkone, line 96: I wonder why notifying the agents precedes updating the db. I can see that this might cause an error, which could prevent the database from being updated.

src/graphql/node/control.rs
sophie-cluml: Done

src/graphql/node/control.rs
sophie-cluml: In a separate meeting, the following points were synced on this:

- Conclusion: We will update the database first and then notify the agents. We will not perform DB rollback actions within the `applyNode` API function due to the failures in agent notification.
- Background for the Conclusion
  - It was previously agreed that "An agent provides a function to indicate whether it is running properly without any failures during the (re)loading of an (updated) configuration."
  - In this context, "running properly" encompasses not only the agent being operational (or alive) but also whether it is functioning with the expected configuration values. Since verifying if the agents are running with the correct configurations requires a separate function that checks and compares the expected and actual configurations, a rollback decision within `applyNode` would not be reliable.
  - Moreover, the `applyNode` API will return a response based on the success or failure of the database update, independent of whether agent notifications succeed.

CHANGELOG.md
```diff
@@ -68,6 +70,25 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added a `language` field to the `Account`. Consequently, the `account` and
   `accountList` API responses now include this field. The `insertAccount` and
   `updateAccount` GraphQL API endpoints are also updated to support the field.
+- Updated the `applyNode` GraphQL API to align with the new node and agent
+  management approach.
+  - The API notifies agents to update their configurations, updates the database
+    with draft values, and logs the changes, as long as each step is needed.
+  - The `successModules` field has been removed from the API response. Instead,
+    the response now includes `gigantoDraft`, representing the draft
+    configuration of "giganto" module. If `gigantoDraft` is `None`, it means
```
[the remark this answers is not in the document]
sophie-cluml: Done

src/graphql/node/control.rs
```diff
@@ -39,172 +38,233 @@ impl NodeControlMutation {
         }
     }
 
-    // TODO: Apply node issue #251
+    /// Applies the draft configuration to the node with the given ID.
+    ///
+    /// This function updates the node's `name` with `name_draft`, `profile` with `profile_draft`,
+    /// and `config` values of agents with their `draft` values. The function returns success as
+    /// long as the database update is successful, regardless of the outcome of notifying agents or
```
sehkone, line 45: I suggest putting the following:

... draft values.

Returns success as long as ...

src/graphql/node/control.rs
```diff
@@ -39,172 +38,233 @@ impl NodeControlMutation {
...
-            &profile_draft.hostname,
-            module_name.as_ref(),
-            &config,
-        )
-        .await?
-        {
-            result_combined.push(module_name);
-        }
-    }
+async fn notify_agents(
```
sehkone, line 142: I don't think the name `notify_agents` is quite fitting because this function not only notifies the agents but also sends the new configuration and updates them.

src/graphql/node/control.rs
```diff
@@ -39,172 +38,233 @@ impl NodeControlMutation {
...
-    module_name: &str,
-    config: &review_protocol::types::Config,
-) -> anyhow::Result<bool> {
-    for _ in 0..MAX_SET_CONFIG_TRY_COUNT {
-        let set_config_response = agents.set_config(hostname, module_name, config).await;
-        if set_config_response.is_ok() {
-            return Ok(true);
-        }
-        info!("Failed to set config for module {module_name}. Retrying...");
+    let notification_results: Vec<Result<_>> = join_all(update_futures).await;
```
sehkone, line 159: Accordingly, it would be better to change `notification_results` to a more appropriate name.

src/graphql/node/control.rs
```diff
@@ -39,172 +38,233 @@ impl NodeControlMutation {
...
     }
+}
+
+fn valid_node_apply(node: &Node) -> bool {
+    // check validity of node itself for apply
+    let valid_name_draft = node.name_draft.is_some();
+    let valid_profile_draft = node.profile_draft.is_some();
 
-    Ok(false)
+    valid_name_draft && valid_profile_draft
```
sehkone, line 178: Wouldn't it be Okay to have `None` for `profile_draft`?

src/graphql/node/control.rs
```diff
@@ -39,172 +38,233 @@ impl NodeControlMutation {
...
-    let old: review_database::NodeUpdate = node.into();
-    let new: review_database::NodeUpdate = updated_node.into();
-    Ok(map.update(i, &old, &new)?)
-}
+    let mut update = node.clone();
+    update.name = if let Some(ref name_draft) = update.name_draft {
+        name_draft.clone()
+    } else {
+        return Err("Name draft must exist".into());
+    };
```
sehkone, line 231: I think we might prefer using `ok_or` here instead of the if else clause.

src/backend.rs
```diff
@@ -61,13 +61,8 @@ pub trait AgentManager: Send + Sync {
...
     async fn reboot(&self, _hostname: &str) -> Result<(), anyhow::Error>;
 
-    /// Sets the configuration of the given agent.
-    async fn set_config(
-        &self,
-        _hostname: &str,
-        _agent_id: &str,
-        _config: &Config,
-    ) -> Result<(), anyhow::Error>;
+    /// Notifies agents to update their configuration.
```
sehkone, line 64: I don't think the term "notify" is exactly what this function does.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   during sign-in.
 - Added ip access control based on the `allow_access_from` field of `Account`
   during sign-in.
+- Added `AgentManager::update_config` method to notify agents to update their
+  configurations.
```
sehkone, line 20: I don't think simply using the term "notify" would suffice because we not only notify agents but also send the new configuration.

sehkone: I was wrong. The manager notifies the agents, and they fetch the configuration. Sorry for the disturbance.

src/backend.rs
sehkone: I was wrong as the above.

src/graphql/node/control.rs
sehkone: I was wrong as the above.

src/graphql/node/control.rs
sehkone: I was wrong as the above.

src/graphql/node/control.rs
sehkone: You've changed this part according to my suggestion.

src/graphql/node/control.rs
```diff
@@ -39,172 +38,230 @@ impl NodeControlMutation {
         }
     }
 
-    // TODO: Apply node issue #251
+    /// Applies the draft configuration to the node with the given ID.
+    ///
+    /// This function updates the node's `name` with `name_draft`, `profile` with `profile_draft`,
+    /// and `config` values of agents with their `draft` values.
+    ///
+    /// Success is returned as long as the database update is successful, regardless of the outcome
```
sehkone, line 46: As I commented, it would be more conventional to start with `Returns` as the first line starts with `Applies`. I attached an example.

스크린샷 2024-09-02 오후 8 48 09

src/graphql/node/control.rs
sophie-cluml: Done

sehkone (approved):

PR #283, opened 2024-08-26
Use `f64` for `ping` field in `NodeStatus`

### LINKED ISSUE #282: Standardize `ping` field in `NodeStatus` GraphQL type
## Current Behavior
Currently, the `ping` field in the `NodeStatus` GraphQL type is represented as a `StringNumber`, which is a custom scalar type.

## Proposed Change
Change the `ping` field to return a `Float` scalar type instead of `StringNumber`.

## Rationale
We aim to minimize the use of custom scalar types like `StringNumber`, by using standard GraphQL types as much as possible. This change will improve compatibility with standard GraphQL clients.

## Additional Notes
- Client-side code that consumes this field may need to be updated to handle a `Float` instead of a `StringNumber`.

msk:
This PR changes the `ping` field in `NodeStatus` in GraphQL API from `StringNumber`, a custom scalar type, to `Float`, a built-in scalar type.

Resolves #282.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Refactor `AgentManager::ping` to return `Duration` instead of `i64`. This
   refactor improves the flexibility and accuracy of the `ping` method, making it
   more robust and aligned with Rust's time handling conventions.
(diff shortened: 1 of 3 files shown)


sophie-cluml (approved):
LGTM

kimhanbeom (approved):

PR #291, opened 2024-08-29
Add signInWithNewPassword GraphQL API

### LINKED ISSUE #14: 첫 로그인 시 계정 비밀번호 변경을 위한 새로운 GraphQL mutation 구현
**이슈 설명:**

계정 보안을 강화하는 일환으로, 모든 계정이 처음으로 로그인 시 비밀번호를 변경하도록 할 필요가 있습니다. 이렇게 하면 디폴트 비밀번호가 첫 로그인 즉시 교체되므로 남용될 위험이 최소화됩니다.

이 요구사항을 구현하기 위해, 새로운 GraphQL mutation인 `sign_in_with_new_password(username: !String, oldPassword: !String, newPassword: !String)`를 추가합니다. 이 mutation은 사용자의 `username`과 `oldPassword`를 사용하여 사용자를 인증하고, 데이터베이스에 저장된 `Account`의 `password` 필드 값을 `newPassword`로 바꾸고 `last_signin_time` 필드를 갱신해야 합니다. `oldPassword`와 `newPassword`는 서로 달라야 합니다.

또한, 계정이 한 번도 로그인한 적이 없는 경우 로그인 시도를 받아들이지 않도록 기존의 `signIn` mutation을 수정해야 합니다.

**구현 전:**

`signIn` mutation은 사용자가 이전에 로그인한 적이 없더라도 로그인 시도를 받아들입니다.

**구현 후:**

1. 첫 사용자가 로그인을 시도합니다.
2. `signIn` mutation은 `last_signin_time`이 `None`인 경우 로그인 시도를 거부합니다.
3. 사용자는 `sign_in_with_new_password`를 사용하여 로그인하고 동시에 비밀번호를 변경합니다.
4. `sign_in_with_new_password` mutation은 사용자를 인증하고, `password`와 `last_signin_time`을 업데이트하고, `AuthPayload`를 리턴합니다.

요청인: @sehkone

henry0715-dev:
Close #14

Changed `signIn` GraphQL API logic.
- Returns `Err` if `last_signin_time` of `account` is `None`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 
 - Added the `TimeSeriesGenerator` variant to the `AgentKind` enum.
+- Added `signInWithNewPassword` GraphQL API for signing in with a new password.
+  - The `signIn` GraphQL API now returns an error if the account was never
+    signed in before. This change is part of enhancing account security by
+    requiring all users to update their passwords upon their first signing in.
 
 ### Changed
 
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -12,7 +12,7 @@ use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use review_database::{
     self as database,
     types::{self},
-    Direction, Iterable, Store,
+    Direction, Iterable, Store, Table,
 };
 use serde::Serialize;
 use tracing::info;
@@ -302,7 +302,16 @@ impl AccountMutation {
         Ok(username)
     }
 
-    /// Authenticates with the given username and password
+    /// Authenticates with the given username and password.
+    ///
+    /// If the `lastSigninTime` value of the `account` is `None`, the operation will fail, and
+    /// it should be guided to call `signInWithNewPassword` GraphQL API.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if the password is invalid, this is the first sign-in attempt, the access
+    /// doesn't originate from a permitted IP address, or the number of sessions exceeds the
+    /// maximum limit.
     async fn sign_in(
         &self,
         ctx: &Context<'_>,
@@ -314,62 +323,45 @@ impl AccountMutation {
         let client_ip = get_client_ip(ctx);
 
         if let Some(mut account) = account_map.get(&username)? {
-            if let Some(allow_access_from) = account.allow_access_from.as_ref() {
-                if let Some(socket) = client_ip {
-                    let ip = socket.ip();
-                    if !allow_access_from.contains(&ip) {
-                        info!("access denied for {username} from IP: {ip}");
-                        return Err("access denied from this IP".into());
-                    }
-                } else {
-                    info!("unable to retrieve client IP for {username}");
-                    return Err("unable to retrieve client IP".into());
-                }
-            }
+            validate_password(&account, &username, &password)?;
+            validate_last_signin_time(&account, &username)?;
+            validate_allow_access_from(&account, &client_ip, &username)?;
+            validate_max_parallel_sessions(&account, &store, &username)?;
 
-            if account.verify_password(&password) {
-                if let Some(max_parallel_sessions) = account.max_parallel_sessions {
-                    let access_token_map = store.access_token_map();
-                    let count = access_token_map
-                        .iter(Direction::Forward, Some(username.as_bytes()))
-                        .filter_map(|res| {
-                            if let Ok(access_token) = res {
-                                if access_token.username == username {
-                                    Some(access_token)
-                                } else {
-                                    None
-                                }
-                            } else {
-                                None
-                            }
-                        })
-                        .count();
-                    if count >= max_parallel_sessions as usize {
-                        info!("maximum parallel sessions exceeded for {username}");
-                        return Err("maximum parallel sessions exceeded".into());
-                    }
-                }
+            sign_in_actions(&mut account, &store, &account_map, &client_ip, &username)
+        } else {
+            info!("{username} is not a valid username");
+            Err("incorrect username or password".into())
+        }
+    }
 
-                let (token, expiration_time) =
-                    create_token(account.username.clone(), account.role.to_string())?;
-                account.update_last_signin_time();
-                account_map.put(&account)?;
+    /// Authenticates with the given username and password, then updates to the new password.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if the password or the new password are invalid, the access
+    /// doesn't originate from a permitted IP address, or the number of sessions exceeds the
+    /// maximum limit.
+    async fn sign_in_with_new_password(
+        &self,
+        ctx: &Context<'_>,
+        username: String,
+        password: String,
+        new_password: String,
+    ) -> Result<AuthPayload> {
+        let store = crate::graphql::get_store(ctx).await?;
+        let account_map = store.account_map();
+        let client_ip = get_client_ip(ctx);
 
-                insert_token(&store, &token, &username)?;
+        if let Some(mut account) = account_map.get(&username)? {
+            validate_password(&account, &username, &password)?;
+            validate_allow_access_from(&account, &client_ip, &username)?;
+            validate_max_parallel_sessions(&account, &store, &username)?;
+            validate_update_new_password(&password, &new_password, &username)?;
 
-                if let Some(socket) = client_ip {
-                    info!("{username} signed in from IP: {}", socket.ip());
-                } else {
-                    info!("{username} signed in");
-                }
-                Ok(AuthPayload {
-                    token,
-                    expiration_time,
-                })
-            } else {
-                info!("wrong password for {username}");
-                Err("incorrect username or password".into())
-            }
+            account.update_password(&new_password)?;
+
+            sign_in_actions(&mut account, &store, &account_map, &client_ip, &username)
         } else {
             info!("{username} is not a valid username");
             Err("incorrect username or password".into())
@@ -462,6 +454,104 @@ impl AccountMutation {
 
         Ok(new_language)
     }
+}
+
+fn validate_password(account: &types::Account, username: &str, password: &str) -> Result<()> {
+    if !account.verify_password(password) {
+        info!("wrong password for {username}");
+        return Err("incorrect username or password".into());
+    }
+    Ok(())
+}
+
+fn validate_last_signin_time(account: &types::Account, username: &str) -> Result<()> {
+    if account.last_signin_time().is_none() {
+        info!("a password change is required to proceed for {username}");
+        return Err("a password change is required to proceed".into());
+    }
+    Ok(())
+}
+
+fn validate_allow_access_from(
+    account: &types::Account,
+    client_ip: &Option<SocketAddr>,
+    username: &str,
+) -> Result<()> {
+    if let Some(allow_access_from) = account.allow_access_from.as_ref() {
+        if let Some(socket) = client_ip {
+            let ip = socket.ip();
+            if !allow_access_from.contains(&ip) {
+                info!("access denied for {username} from IP: {ip}");
+                return Err("access denied from this IP".into());
+            }
+        } else {
+            info!("unable to retrieve client IP for {username}");
+            return Err("unable to retrieve client IP".into());
+        }
+    }
+    Ok(())
+}
+
+fn validate_max_parallel_sessions(
+    account: &types::Account,
+    store: &Store,
+    username: &str,
+) -> Result<()> {
+    if let Some(max_parallel_sessions) = account.max_parallel_sessions {
+        let access_token_map = store.access_token_map();
+        let count = access_token_map
+            .iter(Direction::Forward, Some(username.as_bytes()))
+            .filter_map(|res| {
+                if let Ok(access_token) = res {
+                    if access_token.username == username {
+                        Some(access_token)
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .count();
+        if count >= max_parallel_sessions as usize {
+            info!("maximum parallel sessions exceeded for {username}");
+            return Err("maximum parallel sessions exceeded".into());
+        }
+    }
+    Ok(())
+}
+
+fn validate_update_new_password(password: &str, new_password: &str, username: &str) -> Result<()> {
+    if password.eq(new_password) {
+        info!("password is the same as the previous one for {username}");
+        return Err("password is the same as the previous one".into());
+    }
+    Ok(())
+}
+
+fn sign_in_actions(
+    account: &mut types::Account,
+    store: &Store,
+    account_map: &Table<types::Account>,
+    client_ip: &Option<SocketAddr>,
+    username: &str,
+) -> Result<AuthPayload> {
+    let (token, expiration_time) =
+        create_token(account.username.clone(), account.role.to_string())?;
+    account.update_last_signin_time();
+    account_map.put(account)?;
+
+    insert_token(store, &token, username)?;
+
+    if let Some(socket) = client_ip {
+        info!("{username} signed in from IP: {}", socket.ip());
+    } else {
+        info!("{username} signed in");
+    }
+    Ok(AuthPayload {
+        token,
+        expiration_time,
+    })
 }
 
 /// Returns the expiration time according to the account policy.
@@ -677,7 +767,7 @@ pub fn reset_admin_password(store: &Store) -> anyhow::Result<()> {
 fn initial_credential() -> anyhow::Result<types::Account> {
     let (username, password) = read_review_admin()?;
 
-    let initial_account = review_database::types::Account::new(
+    let initial_account = types::Account::new(
         &username,
         &password,
         database::Role::SystemAdministrator,
@@ -728,6 +818,14 @@ mod tests {
         account::{read_review_admin, REVIEW_ADMIN},
         BoxedAgentManager, MockAgentManager, RoleGuard, TestSchema,
     };
+
+    async fn update_account_last_signin_time(schema: &TestSchema, name: &str) {
+        let store = schema.store().await;
+        let map = store.account_map();
+        let mut account = map.get(name).unwrap().unwrap();
+        account.update_last_signin_time();
+        let _ = map.put(&account).is_ok();
+    }
 
     #[tokio::test]
     #[serial]
@@ -1055,6 +1153,7 @@ mod tests {
         assert_eq!(env::var(REVIEW_ADMIN), Ok("admin:admin".to_string()));
 
         let schema = TestSchema::new().await;
+        update_account_last_signin_time(&schema, "admin").await;
         let res = schema
             .execute(
                 r#"mutation {
@@ -1358,6 +1457,8 @@ mod tests {
             .await;
         assert_eq!(res.data.to_string(), r#"{insertAccount: "u1"}"#);
 
+        update_account_last_signin_time(&schema, "u1").await;
+
         let res = schema
             .execute(
                 r#"mutation {
@@ -1433,6 +1534,8 @@ mod tests {
             .await;
         assert_eq!(res.data.to_string(), r#"{insertAccount: "u1"}"#);
 
+        update_account_last_signin_time(&schema, "u1").await;
+
         let res = schema
             .execute(
                 r#"mutation {
@@ -1467,6 +1570,8 @@ mod tests {
             )
             .await;
         assert_eq!(res.data.to_string(), r#"{insertAccount: "u1"}"#);
+
+        update_account_last_signin_time(&schema, "u1").await;
 
         let res = schema
             .execute(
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -68,6 +68,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added a `language` field to the `Account`. Consequently, the `account` and
   `accountList` API responses now include this field. The `insertAccount` and
   `updateAccount` GraphQL API endpoints are also updated to support the field.
+- Added `common_sign-in_logic` to handle `sign-in` GraphQL API logic.
```
sehkone, line 71: This is not necessary because the method is not for users.

src/graphql/account.rs
```diff
@@ -420,6 +373,125 @@ impl AccountMutation {
...
+/// - If the number of active sessions exceeds the account's `max_parallel_sessions` limit,
+///   a "maximum parallel sessions exceeded" error is returned.
+/// - If the provided password does not match the stored password, an "incorrect username or
+///   password" error is returned.
+/// - If the `is_update_password` flag is set to true and the new password is the same as the
+///   current password, a "password is the same as the previous one" error is returned.
+/// - If the account's `last_signin_time` is `None` and a password change is required, an
+///   "a password change is required to proceed" error is returned.
+/// - If there is an error creating a token, updating the last sign-in time, or storing the
+///   account or token data, the respective error messages are returned.
```
sehkone, line 404: Internal methods do not require the Rust documentation. Explanation suffices, if needed.

src/graphql/account.rs
```diff
@@ -287,78 +287,31 @@ impl AccountMutation {
         Ok(username)
     }
 
-    /// Authenticates with the given username and password
+    /// - Authenticates with the given username and password.
+    /// - If the `last_signin_time` value of the `account` is `None`, the operation will fail, and
+    /// it should be guided to call common_sign_in_logic.
```
sehkone, line 292: This method is open to users. It means the Rust documentation should be fulfilled according to [the convention](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html).

CHANGELOG.md
henry0715-dev: I'll delete it.

src/graphql/account.rs
sehkone, line 292: Since this is a open method to users, we need to fulfill the Rust documentation according to [the convention](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html)

src/graphql/account.rs
```diff
@@ -287,78 +287,58 @@ impl AccountMutation {
...
+    ///
+    /// ```ignore
+    /// r#"mutation {
+    ///     signInWithNewPassword(username: "user", password: "pw", newPassword: "pw2") {
+    ///         token
+    ///     }
+    /// }"#
+    /// ```
+    ///
+    /// # Returns
```
sehkone, line 306: We usually don't mention returning values in a separate section. You can refer to the standard library to check how to write the Rust documentation.

src/graphql/account.rs
```diff
@@ -420,6 +400,96 @@ impl AccountMutation {
     }
 }
 
+async fn common_sign_in_logic(
```
sehkone, line 403: This method doesn't seem to behave consistently. Its behavior depends on `is_update_password`, which is the only differing argument between the two callers. In other words, this method is not the common part between the two callers. When we extract the common part, it should be identical across all participants.

sehkone: If there were any reason to keep this, it should have at least a different name rather than "common".

henry0715-dev: What do you think of the function name below?
```
sign_in_logic_by_update_password
```

src/graphql/account.rs
henry0715-dev: I'll delete it.

src/graphql/account.rs
sehkone: It would be better to explicitly outline the process in both methods, `sign_in` and `sign_in_with_new_password`. 

I suggest the following pseudocode:

`sign_in`
    - Check the password
    - Check if there was no previous sign-in
    - Check the IP address
    - Check the number of parallel sessions
    - Sign in

`sign_in_with_new_password`
    - Check the password
    - Check the IP address
    - Check the number of parallel sessions
    - Change the password
    - Sign in

While revealing this skeleton in both methods, we need to extract the common parts as separate methods.

sehkone: I want to point out we need to check the password first. The information about whether the user is signing in for the first time or again, whether the number of sessions exceeds the limit, or whether the IP address is valid, is so confidential that it should not be accessible to anyone who has not provided the correct password. The original code was not properly implemented in this regard.

sehkone: > While revealing this skeleton in both methods, we need to extract the common parts as separate methods.

The common part might involve validating the IP address and the number of sessions.

henry0715-dev: I considered splitting the function because it is responsible for so many functions, but there is a lot of overlap, so we made it into one function.
As you mentioned above, we will refactor the checking part by splitting it into functions and calling them.

src/graphql/account.rs
```diff
@@ -641,6 +742,7 @@ fn initial_credential() -> anyhow::Result<types::Account> {
         None,
         None,
     )?;
+    initial_account.update_last_signin_time();
```
sehkone, line 745: Could you explain why you put this?

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added a `language` field to the `Account`. Consequently, the `account` and
   `accountList` API responses now include this field. The `insertAccount` and
   `updateAccount` GraphQL API endpoints are also updated to support the field.
+- Added `signInWithNewPassword` GraphQL API for signing in with a new password.
```
sehkone, line 71: This is a new API, so should be included in the section "Added".

sehkone: Since the `signIn` has been changed, it should be documented.

src/graphql/account.rs
henry0715-dev: This is the part set to satisfy the existing test case.
There is also a part where admin/admin does not affect the part you are currently using when running for the first time.
In the future, I thought of the expected behavior of creating a new account after logging in with the admin account, and requesting password change when attempting to log in with the created account.

CHANGELOG.md
henry0715-dev: I'll change it.

CHANGELOG.md
```diff
@@ -68,6 +69,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Added a `language` field to the `Account`. Consequently, the `account` and
   `accountList` API responses now include this field. The `insertAccount` and
   `updateAccount` GraphQL API endpoints are also updated to support the field.
+- Added logic to check the `lastSigninTime` value in the `signIn` GraphQL API.
```
sehkone, line 72: I think we need to document how the new logic works, as well as the fact that it has changed, because in this case, the new behavior is the primary concern for users.

src/graphql/account.rs
sehkone: I don't quite understand. Could you elaborate on what you answered?

sehkone: This `initial_credential` seems to create the first admin account and immediately set its sign-in time. This makes it pointless to alter `sign_in` to check the last sign-in time. What do you think?

src/graphql/account.rs
sehkone: When responding to a reviewer's request for changes, it's better to do so both when you first review the request and when you've completed the requested changes . If I had to choose one, I would prefer the latter.

CHANGELOG.md
henry0715-dev: Added content.

src/graphql/account.rs
henry0715-dev: Yes, I see.
The function that left document has been removed, making the long document above completely unnecessary.

CHANGELOG.md
sehkone: The key point is that this API rejects a user's access if it is their first attempt. Therefore, it's not enough to simply mention checking the `lastSigninTime`. Additionally, the word "logic" seems unnecessary. I suggest:

- Changed the `signIn` to reject the first sign-in attempt unless it is done through the new `signInWithNewPassword` API.

In this case, "Changed" or "Updated" is preferred over "Added" because the behavior of the API has actually changed from the user's perspective. Also, the word "reject" or any similar one should be included to convey the key point.

sehkone:
When reseting the admin's password, we need to reset the `last_signin_time` as well.

CHANGELOG.md
henry0715-dev: I will delete the explanation below and change it to the sentence you suggested.
**"Apply a requirement for a password change upon the initial `signIn`."**

src/graphql/account.rs
henry0715-dev: This part has been restored to its original state.

src/graphql/account.rs
```diff
@@ -301,7 +301,25 @@ impl AccountMutation {
...
+    /// If the `signIn` GraphQL API response is the error message "a password change is
+    /// required to proceed", you can call the `signInWithNewPassword` GraphQL API.
+    ///
+    /// ```ignore
+    /// r#"mutation {
+    ///     signInWithNewPassword(username: "user", password: "pw", newPassword: "pw2") {
+    ///         token
+    ///     }
+    /// }"#
+    /// ```
```
sehkone, line 318: We don't need to present the signatures of GraphQL APIs in the code because users can not only infer them on their own but also check them in the GraphQL playground.

src/graphql/account.rs
```diff
@@ -301,7 +301,25 @@ impl AccountMutation {
...
     }
 
-    /// Authenticates with the given username and password
+    /// Authenticates with the given username and password.
+    ///
+    /// If the `lastSigninTime` value of the `account` is `None`, the operation will fail, and
+    /// it should be guided to call `signInWithNewPassword` GraphQL API.
+    ///
+    /// If the `signIn` GraphQL API response is the error message "a password change is
+    /// required to proceed", you can call the `signInWithNewPassword` GraphQL API.
```
sehkone, line 310: I don't think this part is necessary because it seems to repeat what was mentioned right above.

src/graphql/account.rs
```diff
@@ -301,7 +301,25 @@ impl AccountMutation {
...
+    /// r#"mutation {
+    ///     signInWithNewPassword(username: "user", password: "pw", newPassword: "pw2") {
+    ///         token
+    ///     }
+    /// }"#
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// An error message (`String`) if any of the checks fail or an error occurs during the process.
```
sehkone, line 322: I suggest:

```rust
/// # Errors
///
/// Returns `Err` if the password is invalid, this is the first sign-in attempt, the access doesn't originate from a permitted IP address, or the number of sessions exceeds the maximum limit.
```

The style where the sentence starts with a third-person singular verb is a convention in Rust documentation. Additionally, it would be better to specify exactly the cases where errors occur.

sophie-cluml:
@sehkone I know we are still in pr-review stage for this PR, but I think merging PR would require implement new UX flow and design in the client side. So I would like to suggest in advance, that we might merge this PR after we release the upcoming review-web version.

src/graphql/account.rs
sehkone: If you deleted that part, that’s not what I meant. I just wanted to say that the return values should be mentioned according to the Rust documentation convention, like this:

```rust
///
/// Returns blah blah blah
///
```

- The title, `# Returns`, is not necessary.
- The sentence starts with `Returns`, a third-person singular verb.

src/graphql/account.rs
sehkone: The part I think is unnecessary:

```rust
/// If the `signIn` GraphQL API response is the error message "a password change is
/// required to proceed", you can call the `signInWithNewPassword` GraphQL API.
```

sehkone:
> @sehkone I know we are still in pr-review stage for this PR, but I think merging PR would require implement new UX flow and design in the client side. So I would like to suggest in advance, that we might merge this PR after we release the upcoming review-web version.

Yes. I won't merge this until what is required is done.

src/graphql/account.rs
henry0715-dev: I'll delete it.

src/graphql/account.rs
henry0715-dev: I'll change it.

src/graphql/account.rs
henry0715-dev: I'll delete it.

src/graphql/account.rs
henry0715-dev: I'll be careful.

src/graphql/account.rs
```diff
@@ -301,7 +301,15 @@ impl AccountMutation {
         Ok(username)
     }
 
-    /// Authenticates with the given username and password
+    /// Authenticates with the given username and password.
+    ///
+    /// If the `lastSigninTime` value of the `account` is `None`, the operation will fail, and
+    /// it should be guided to call `signInWithNewPassword` GraphQL API.
+    ///
+    /// # Errors
```
sehkone, line 309: It would be better to insert an empty line between the title and the content.

henry0715-dev: Thank you for checking the mistake.

sehkone:
@msk 

This PR hasn't been merged yet because it requires the corresponding feature to be added to the UI. Without this feature, users essentially cannot sign in the system.

In the meantime, since requiring users to change their password at the first sign-in is still inconvenient, I think it would be better to make this optional based on a command line option or another approach. Could you suggest your idea on this?

msk:
> This PR hasn't been merged yet because it requires the corresponding feature to be added to the UI. Without this feature, users essentially cannot sign in the system.

I acknowledge that the PR cannot be merged until the corresponding UI feature is in place.

> In the meantime, since requiring users to change their password at the first sign-in is still inconvenient, I think it would be better to make this optional based on a command line option or another approach. Could you suggest your idea on this?

I believe it would be more appropriate to discuss potential changes to the original issue (#14) there rather than within the PR itself. I’ll move this part of the discussion to the issue thread.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+ 
+- Added `signInWithNewPassword` GraphQL API for signing in with a new password.
+
```
sehkone, line 13: @henry0715-dev,

Could you take a look at this? It indicates that unnecessary spaces are included, which suggests you may need to check your editor setup, as VS Code shouldn't have done this if it were set up properly.

henry0715-dev: I've completed the modification.
Thank you.

src/graphql/account.rs
```diff
@@ -1593,4 +1698,179 @@ mod tests {
...
+        let query = r#"mutation {
+                    signIn(username: "u1", password: "pw2") {
+                        token
+                    }
+              }"#;
+        let res = schema.execute(query).await;
+        assert_eq!(
+            res.errors.first().unwrap().message.to_string(),
+            "incorrect username or password".to_string()
+        );
```
sophie-cluml, line 1747: Could you please clarify the goals of this section of the test code?

From what I understand, lines 1704-1731 seem to be validating that a password change is required to proceed at first sign-in. However, I'm not entirely certain about the purpose of lines 1733-1747. Specifically, I’m curious about why `update_account_last_signin_time` is called and why an incorrect username is passed to the signIn GraphQL API. Were you aiming to conduct a general test of the `signIn` function with these steps? If so, I'd appreciate if you could come up with more general naming for the test code function. If you were aiming just to make sure that users cannot proceed without the password change, I think we can just get rid of L1733-1747, since I think `password_validate_proceed` test code targets to cover that scenario.

sophie-cluml (commented):
I'd like to ask you for a rebase please.

sophie-cluml:
Thank you for your kindness in describing your PR.  I would like to suggest a small update to align with our convention. Please consider changing

> Changed `sign-in` GraphQL API logic.

to

- Changed `signIn` GraphQL API logic.

src/graphql/account.rs
```diff
@@ -464,6 +456,104 @@ impl AccountMutation {
...
 
+fn validate_password(account: &types::Account, username: &str, password: &str) -> Result<()> {
+    if !account.verify_password(password) {
+        info!("wrong password for {username}");
+        return Err("incorrect username or password".into());
+    }
+    Ok(())
+}
+
+fn validate_last_signin_time(account: &types::Account, username: &str) -> Result<()> {
```
sophie-cluml, line 467: I would like to suggest using `validate_last_sign_in_time` instead of `validate_last_signin_time` for consistency. 

From what I googled online, I got this search result, so I am leaning to 'sign in', rather than `signin` or `sign-in`.

image

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added `signInWithNewPassword` GraphQL API for signing in with a new password.
+
 ### Changed
```
sophie-cluml, line 15: In line with this [comment](https://github.com/aicers/review-web/pull/291#issuecomment-2449076264), I noticed that the `signIn` API logic has been modified. As this logic impacts end users, I think it seems beneficial to explicitly include this change in the changelog to keep our users informed. I’d suggest adding a line in the Changed section, perhaps like the following:

- The `signIn` GraphQL API now returns an error if the account was never signed in before. This change is part of enhancing account security by requiring all users to update their passwords upon their first signing in.

src/graphql/account.rs
henry0715-dev: I'll remove it.
Thank you.

src/graphql/account.rs
```diff
@@ -314,62 +323,45 @@ impl AccountMutation {
...
-                    create_token(account.username.clone(), account.role.to_string())?;
-                account.update_last_signin_time();
-                account_map.put(&account)?;
+    /// Authenticates with the given username and password, and update the new password.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` if the password or the new password are invalid, the access
+    /// doesn't originate from a permitted IP address, or the number of sessions exceeds the
+    /// maximum  limit.
```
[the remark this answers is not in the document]
henry0715-dev: I'll change it.

sehkone:
> I'd like to ask you for a rebase please.

@sophie-cluml,

I don't see any conflicts with the main branch. Why would you think this needs a rebase?

sehkone:
> Thank you for your kindness in describing your PR. I would like to suggest a small update to align with our convention. Please consider changing
> 
> > Changed `sign-in` GraphQL API logic.
> 
> to
> 
> * Changed `signIn` GraphQL API logic.

The representation, such as `signIn`, is not merely conventional. Since it is defined in the schema, we must use it as is.

sehkone:
> > Thank you for your kindness in describing your PR. I would like to suggest a small update to align with our convention. Please consider changing
> > > Changed `sign-in` GraphQL API logic.
> > 
> > 
> > to
> > 
> > * Changed `signIn` GraphQL API logic.
> 
> The representation, such as `signIn`, is not merely conventional. Since it is defined in the schema, we must use it as is.

One more thing: backticks should be used to quote elements that belong to another context. When we search for a quoted string in the original text, it must appear exactly as quoted. There is no `sign-in` in our codebase. Across the code and the UI, we only have two representations: `sign_in` and `signIn`.

henry0715-dev:
> > I'd like to ask you for a rebase please.
> 
> @sophie-cluml,
> 
> I don't see any conflicts with the main branch. Why would you think this needs a rebase?

When I commented, it was conflicting.
Currently, rebase is applied.

sehkone:
> > > > Thank you for your kindness in describing your PR. I would like to suggest a small update to align with our convention. Please consider changing
> > > > > Changed `sign-in` GraphQL API logic.
> > > > 
> > > > 
> > > > to
> > > > 
> > > > * Changed `signIn` GraphQL API logic.
> > > 
> > > 
> > > The representation, such as `signIn`, is not merely conventional. Since it is defined in the schema, we must use it as is.
> > 
> > 
> > One more thing: backticks should be used to quote elements that belong to another context. When we search for a quoted string in the original text, it must appear exactly as quoted. There is no `sign-in` in our codebase. Across the code and the UI, we only have two representations: `sign_in` and `signIn`.
> 
> Commit message modification completed. Thank you.

For clarity, I need to augment my previous explanation. Even though backticks are for quoting code elements, they can be used for any string that might or will appear in any code context. Therefore, it doesn't mean they implicate that the string enclosed by them must be searchable. But, here in this context, i.e. CHANGELOG.md, it's assumed that it should be searchable.

To clarify further: while backticks are typically used to quote code elements, they can also be applied to any string that might or will appear in a code context. In other words, backticks doesn’t necessarily imply that the enclosed string must be searchable. However, in this CHANGELOG.md, it’s assumed that it should be searchable.

henry0715-dev:
> > > > > Thank you for your kindness in describing your PR. I would like to suggest a small update to align with our convention. Please consider changing
> > > > > > Changed `sign-in` GraphQL API logic.
> > > > > 
> > > > > 
> > > > > to
> > > > > 
> > > > > * Changed `signIn` GraphQL API logic.
> > > > 
> > > > 
> > > > The representation, such as `signIn`, is not merely conventional. Since it is defined in the schema, we must use it as is.
> > > 
> > > 
> > > One more thing: backticks should be used to quote elements that belong to another context. When we search for a quoted string in the original text, it must appear exactly as quoted. There is no `sign-in` in our codebase. Across the code and the UI, we only have two representations: `sign_in` and `signIn`.
> > 
> > 
> > Commit message modification completed. Thank you.
> 
> To clarify further: while backticks are typically used to quote code elements, they can also be applied to any string that might or will appear in a code context. In other words, backticks doesn’t necessarily imply that the enclosed string must be searchable. However, in this CHANGELOG.md, it’s assumed that it should be searchable.

Thanks for the detailed explanation and guide. 
I will be careful with backticks and naming.

src/graphql/account.rs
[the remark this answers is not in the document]
henry0715-dev: Modification completed.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added `signInWithNewPassword` GraphQL API for signing in with a new password.
+  - The `signIn` GraphQL API now returns an error if the account was never signed
```
sophie-cluml, line 13: I think we need to break the line before the word "signed". Could you check your IDE settings?

henry0715-dev: I checked for modification using the cmd + q command in the VSCode IDE.
thank you.

sophie-cluml (approved):

sophie-cluml:
Could you rebase please?

henry0715-dev:
> Could you rebase please?
It's done.

sophie-cluml (approved):

PR #292, opened 2024-08-30
Update the GraphQL API `nodeStatusList`

### LINKED ISSUE #279: Update `nodeStatusList` API to reflect recent changes in node & agent management
### Background 

I think recent changes in node and agent management within the ecosystem where review-web operates needs to be accompanied with an update to the `nodeStatusList` GraphQL API , along with issues #245 and #251. 

### Proposed Change

Currently, the `nodeStatusList` API _mainly_ returns three types of information with `NodeStatus` struct : 1) node resources, 2) the online/offline status of modules as booleans, and 3) module configurations, which are collected from agents. (Current definition of `NodeStatus` : https://github.com/aicers/review-web/blob/main/src/graphql/node.rs#L305 )

However, due to recent management changes, we now need to provide the status of modules more richly, potentially using an enum data type to represent statuses such as "disabled," "enabled," and "reload failed."

Additionally, we no longer need to provide module configurations in this API endpoint. Previously, configuration-related response fields were included so that GraphQL clients could compare the current configuration with a draft configuration to determine if an `applyNode` API request could be made. With the recent changes in node and agent management, what I understand is that GraphQL clients are now expected to use the `config` and `draft` stored in the central database for this comparison and decision.

So, I propose the following changes:

1. Update the data types of the fields that represent the module's online/offline status to support more than two statuses.
2. Remove the `piglet_config` and `hog_config` fields from the `NodeStatus` struct.
3. Change the data type of `NodeStatus::review` to `bool`, because I think boolean is sufficient to indicate whether the central server is on the node or not.
4. Rename the following fields: `piglet` -> `features` , `reconverge` -> `learner`, `hog` -> `models`
5. ~~Modify the `ping` field to return a result that combines `Agent::status` with the actual ping response from the agent.~~ Modify the `features`, `learner`, and `models` field to represent a status that combines `Agent::status` with the actual online status of agent.

I think below is one of possible implementations, (with comments temporarily removed for brevity):

```rust
struct NodeStatus {
    id: u32,
    name: String,
    cpu_usage: Option<f32>,
    total_memory: Option<u64>,
    used_memory: Option<u64>,
    total_disk_space: Option<u64>,
    used_disk_space: Option<u64>,
    ping: Option<i64>,

    review: bool,

    features: AgentStatus,
    learner: AgentStatus,
    models: AgentStatus,
}

enum AgentStatus {
    Disabled 
    Enabled,
    ReloadFailed,
    Unknown,
}

### PR DESCRIPTION
**Pull request by sophie-cluml**
close #279 

-- 
This commit is worked on top of #280

### DIFF
--- a/src/graphql/node.rs
+++ b/src/graphql/node.rs
@@ -301,13 +301,29 @@ impl From<review_protocol::types::PigletConfig> for PigletConfig {
     }
 }
 
+#[derive(Clone, Deserialize, Serialize, SimpleObject, PartialEq)]
+pub struct AgentSnapshot {
+    kind: AgentKind,
+    stored_status: AgentStatus,
+
+    /// Serialized TOML string containing the configuration of the agent.
+    config: Option<String>,
+
+    /// Serialized TOML string containing the draft configuration of the agent.
+    draft: Option<String>,
+}
+
 #[derive(Clone, Deserialize, Serialize, SimpleObject)]
 #[graphql(complex)]
 pub(super) struct NodeStatus {
     #[graphql(skip)]
     id: u32,
 
-    /// The hostname of the node.
+    /// The name of the node.
+    ///
+    /// This is the user-defined name for the node, which may or may not be the same as the node's
+    /// hostname. The name should not be confused with the node's hostname, even though they could
+    /// be identical by coincidence.
     name: String,
 
     /// The average CPU usage in percent.
@@ -333,23 +349,12 @@ pub(super) struct NodeStatus {
     #[graphql(skip)]
     ping: Option<Duration>,
 
-    /// Whether review is online or not.
-    review: Option<bool>,
+    /// Indicates whether the Manager server is running on this node or not.
+    manager: bool,
 
-    /// Whether piglet is online or not.
-    piglet: Option<bool>,
-
-    /// actual piglet configuration
-    piglet_config: Option<PigletConfig>,
-
-    /// Whether reconverge is online or not.
-    reconverge: Option<bool>,
-
-    /// Whether hog is online or not.
-    hog: Option<bool>,
-
-    /// actual hog configuration
-    hog_config: Option<HogConfig>,
+    /// The list of agents running on the node. `AgentSnapshot` contains the agent's kind, stored
+    /// status in the database, and config and draft configurations.
+    agents: Vec<AgentSnapshot>,
 }
 
 #[ComplexObject]
@@ -357,19 +362,20 @@ impl NodeStatus {
     async fn id(&self) -> ID {
         ID(self.id.to_string())
     }
-    /// The RAM size in bytes within the range representable by a `u64`
+    /// The RAM size in bytes within the range representable by a `u64`.
     async fn total_memory(&self) -> Option<StringNumber<u64>> {
         self.total_memory.map(StringNumber)
     }
-    /// The amount of used RAM in bytes within the range representable by a `u64`
+    /// The amount of used RAM in bytes within the range representable by a `u64`.
     async fn used_memory(&self) -> Option<StringNumber<u64>> {
         self.used_memory.map(StringNumber)
     }
-    /// The total disk space in bytes within the range representable by a `u64`
+    /// The total disk space in bytes within the range representable by a `u64`.
     async fn total_disk_space(&self) -> Option<StringNumber<u64>> {
         self.total_disk_space.map(StringNumber)
     }
-    /// The total disk space in bytes that is currently used within the range representable by a `u64`
+    /// The total disk space in bytes that is currently used within the range representable by a
+    /// `u64`.
     async fn used_disk_space(&self) -> Option<StringNumber<u64>> {
         self.used_disk_space.map(StringNumber)
     }
@@ -390,12 +396,8 @@ impl NodeStatus {
         total_disk_space: Option<u64>,
         used_disk_space: Option<u64>,
         ping: Option<Duration>,
-        review: Option<bool>,
-        piglet: Option<bool>,
-        piglet_config: Option<PigletConfig>,
-        reconverge: Option<bool>,
-        hog: Option<bool>,
-        hog_config: Option<HogConfig>,
+        manager: bool,
+        agents: Vec<AgentSnapshot>,
     ) -> Self {
         Self {
             id,
@@ -406,12 +408,8 @@ impl NodeStatus {
             total_disk_space,
             used_disk_space,
             ping,
-            review,
-            piglet,
-            piglet_config,
-            reconverge,
-            hog,
-            hog_config,
+            manager,
+            agents,
         }
     }
 }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/graphql/node/control.rs [hunk 1]
```diff
@@ -39,188 +38,249 @@ impl NodeControlMutation {
         }
     }
 
-    // TODO: Apply node issue #251
+    /// Applies the draft configuration to the node with the given ID.
+    ///
+    /// This function updates the node's `name` with `name_draft`, `profile` with `profile_draft`,
+    /// and `config` values of agents with their `draft` values. The function returns success as
+    /// long as the database update is successful, regardless of the outcome of notifying agents or
```
sehkone, line 45: I suggest putting the following:

... `draft` values.

Returns success as long as ...

src/graphql/node/control.rs
```diff
@@ -39,188 +38,249 @@ impl NodeControlMutation {
...
-            &profile_draft.hostname,
-            module_name.as_ref(),
-            &config,
-        )
-        .await?
-        {
-            result_combined.push(module_name);
-        }
-    }
+async fn notify_agents(
```
sehkone, line 142: I don't think the name `notify_agents` is quite fitting because this function not only notifies the agents but also sends the new configuration and updates them.

src/graphql/node/control.rs
```diff
@@ -39,188 +38,249 @@ impl NodeControlMutation {
...
-    module_name: &str,
-    config: &review_protocol::types::Config,
-) -> anyhow::Result<bool> {
-    for _ in 0..MAX_SET_CONFIG_TRY_COUNT {
-        let set_config_response = agents.set_config(hostname, module_name, config).await;
-        if set_config_response.is_ok() {
-            return Ok(true);
-        }
-        info!("Failed to set config for module {module_name}. Retrying...");
+    let notification_results: Vec<Result<_>> = join_all(update_futures).await;
```
sehkone, line 159: Accordingly, it would be better to change `notification_results` to a more appropriate name.

src/graphql/node/control.rs
```diff
@@ -39,188 +38,249 @@ impl NodeControlMutation {
...
     }
+}
+
+fn valid_node_apply(node: &Node) -> bool {
+    // check validity of node itself for apply
+    let valid_name_draft = node.name_draft.is_some();
+    let valid_profile_draft = node.profile_draft.is_some();
 
-    Ok(false)
+    valid_name_draft && valid_profile_draft
```
sehkone, line 178: Wouldn't it be Okay to have `None` for `profile_draft`?

src/graphql/node/control.rs
```diff
@@ -39,188 +38,249 @@ impl NodeControlMutation {
...
-    let old: review_database::NodeUpdate = node.into();
-    let new: review_database::NodeUpdate = updated_node.into();
-    Ok(map.update(i, &old, &new)?)
-}
+    let mut update = node.clone();
+    update.name = if let Some(ref name_draft) = update.name_draft {
+        name_draft.clone()
+    } else {
+        return Err("Name draft must exist".into());
+    };
```
sehkone, line 231: I think we might prefer using `ok_or` here instead of the if else clause.

src/graphql/node/status.rs
```diff
@@ -9,12 +9,12 @@ use roxy::ResourceUsage;
...
 use super::{
     super::{BoxedAgentManager, Role, RoleGuard},
-    ModuleName, NodeStatus, NodeStatusQuery, NodeStatusTotalCount,
+    matches_manager_hostname, AgentStatus, NodeStatus, NodeStatusQuery, NodeStatusTotalCount,
 };
 
 #[Object]
 impl NodeStatusQuery {
-    /// A list of status of all the nodes.
+    /// A list of status of nodes.
```
sehkone, line 17: Why would you remove `all the`? Do you think the list that this method returns
might not include all the nodes?

src/graphql/node/status.rs
```diff
@@ -36,195 +36,123 @@ impl NodeStatusQuery {
...
+// Returns the statuses of the Features, Learner, and Models agents for the given hostname.
+fn determine_agent_statuses(
+    online_agents_by_host_id: &HashMap<String, Vec<(String, String)>>,
+    hostname: &str,
+    agents: &[review_database::Agent],
+) -> (AgentStatus, AgentStatus, AgentStatus) {
+    // TODO: The current logic determines an agent's status solely based on its presence in the
+    // `online_agents_by_host_id` map. An agent is marked as `Enabled` if found in the map, and
+    // `Disabled` if not. In the future, we may consider implementing more advanced logic that
+    // compares the agent's status in the DB with its online status.
```
sehkone, line 126: Could you post an issue regarding this instead of leaving a comment? If a
comment is necessary to alert others, we can simply reference the issue number. BTW,
I don't quite understand what "its online status" means, but I think agents should update
their status to the manager whenever they change, which hasn't been implemented yet.

src/graphql/node/control.rs
sehkone: Sorry for the confusion. I should've posted this in the `applyNode` PR.

src/graphql/node/control.rs
sehkone: Sorry for the confusion. I should've posted this in the `applyNode` PR.

src/graphql/node/control.rs
sehkone: Sorry for the confusion. I should've posted this in the `applyNode` PR.

src/graphql/node/control.rs
sehkone: Sorry for the confusion. I should've posted this in the `applyNode` PR.

src/graphql/node/control.rs
sehkone: Sorry for the confusion. I should've posted this in the `applyNode` PR.

src/graphql/node.rs
```diff
@@ -490,8 +484,9 @@ pub enum ModuleName {
     Review,
 }
 
-pub fn is_review(hostname: &str) -> bool {
+pub fn matches_manager_hostname(hostname: &str) -> bool {
+    // Current machine's hostname is the Manager server's hostname, because this code always runs
+    // on the Manager server.
     let review_hostname = roxy::hostname();
-
     !review_hostname.is_empty() && review_hostname == hostname
```
sehkone, line 491: Since you’ve already changed the name of the method, why don't you change this `review` to `manager` as well?

src/graphql/node/status.rs
sophie-cluml: I removed the "all" because this endpoint returns maximum 100 nodes in a single response.

src/graphql/node.rs
sophie-cluml: Done

src/graphql/node/status.rs
sehkone: Thank you for reminding me of that.

src/graphql/node.rs
```diff
@@ -333,23 +345,12 @@ pub(super) struct NodeStatus {
...
 
-    /// Whether reconverge is online or not.
-    reconverge: Option<bool>,
-
-    /// Whether hog is online or not.
-    hog: Option<bool>,
-
-    /// actual hog configuration
-    hog_config: Option<HogConfig>,
+    /// The list of agents running on the node. `AgentData` contains the agent's kind, status, and
```
sehkone, line 351: What does the `AgentData` refer to?

sophie-cluml: It is corrected to `AgentSnapshot`.

sehkone (approved):

PR #307, opened 2024-09-19
Fix `applyNode` GraphQL API

### LINKED ISSUE #303: Improve log message in `applyNode`
I think we don't need "FilterMap" or iterable related info in the log. Below is the log we have at hash 1007233544cc56c25693c12b206ade584e6224fd

```
2024-09-09T04:53:35.699404Z  INFO review_web::graphql::node::control: [2024-09-09 04:53:35.699382 UTC] Node ID 0 - Node's agents are notified to be updated.
FilterMap { iter: Iter([Agent { node: 0, key: "piglet@localhost", kind: Piglet, status: Enabled, config: None, draft: Some(Config { inner: "piglet=1" }) }]) }
```

### LINKED ISSUE #306: Fix Incorrect `agent_key` Format in `AgentManager::update_config` Call in `applyNode`
In the `applyNode` GraphQL API of `review-web`, the `AgentManager::update_config` function is being called with an incorrect format for the `agent_key`. The function `update_config` expects the `agent_key` to be in the format of:

```rust
// note: this is pseudo code
format!("{}@{}", agent.key, node.profile.unwrap().hostname)
```

However, the current implementation only uses `agent.key`, without the "@" sign and the "hostname". So we need to correct it.

sophie-cluml:
close #306
close #303

### DIFF
--- a/src/graphql/node/control.rs
+++ b/src/graphql/node/control.rs
@@ -99,31 +99,38 @@ impl NodeControlMutation {
         }
 
         if let Some(ref target_agents) = apply_scope.agents {
-            let agent_manager = ctx.data::<BoxedAgentManager>()?;
-            if let Err(e) = notify_agents(
-                agent_manager,
-                &target_agents.updates,
-                &target_agents.disables,
-            )
-            .await
-            {
-                error!("Failed to notify agents for node {i} to be updated. However, the failure does not affect the node application operation.\nDetails:\n{e:?}",);
-            }
+            let hostname = {
+                let store = crate::graphql::get_store(ctx).await?;
+                let node_map = store.node_map();
+                let (node, _) = node_map.get_by_id(i)?.ok_or_else(|| {
+                    async_graphql::Error::new(format!("Node with ID {i} not found"))
+                })?;
+                node.profile.map(|p| p.hostname).unwrap_or_default()
+            };
 
-            info!(
-                "[{}] Node ID {i} - Node's agents are notified to be updated.\n{:?}",
-                chrono::Utc::now(),
-                node.agents.iter().filter_map(|agent| {
-                    if target_agents.updates.contains(&agent.key.as_str()) {
-                        Some(format!(
-                            "\nAgent key: {}, Config: {:?}, Draft: {:?}",
-                            agent.key, agent.config, agent.draft
-                        ))
-                    } else {
-                        None
-                    }
-                })
-            );
+            if hostname.is_empty() {
+                info!(
+                    "Node ID {i} - Node's agents are not notified because the hostname is empty."
+                );
+            } else {
+                let agent_manager = ctx.data::<BoxedAgentManager>()?;
+                if let Err(e) = notify_agents(
+                    agent_manager,
+                    hostname.as_str(),
+                    &target_agents.updates,
+                    &target_agents.disables,
+                )
+                .await
+                {
+                    error!("Failed to notify agents for node {i} to be updated. This failure may impact configuration synchronization.\nDetails: {e:?}");
+                }
+
+                info!(
+                    "[{}] Node ID {i} - Node's agents are notified to be updated. {:?}",
+                    chrono::Utc::now(),
+                    target_agents.updates,
+                );
+            }
         }
 
         Ok(ApplyNodeResponse { id, giganto_draft })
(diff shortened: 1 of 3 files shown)


src/graphql/node/control.rs
```diff
@@ -99,37 +99,60 @@ impl NodeControlMutation {
...
-                            agent.key, agent.config, agent.draft
-                        ))
-                    } else {
-                        None
-                    }
-                })
-            );
+            if hostname.is_empty() {
+                info!(
+                    "Node ID {i} - Node's agents are not notified because the hostname is empty."
```
sehkone, line 113: It seems to me that this results in no error being returned to the user, even with this message. Could you explain why?

src/graphql/node/control.rs
```diff
@@ -145,20 +168,25 @@ pub struct ApplyNodeResponse {
...
+            .await
+            .map_err(|e| {
+                async_graphql::Error::new(format!(
+                    "Failed to notify agent for config update {agent_key}: {e}"
+                ))
+            })
     });
 
     // TODO: We need to implement the logic to disable agents. For now, we will only log the
     // message.
```
sehkone, line 188: Could you mention any related issues if any? If none, it would be better to create one, and to mention it here.

src/graphql/node/control.rs
```diff
@@ -99,37 +99,60 @@ impl NodeControlMutation {
...
+                let agent_manager = ctx.data::<BoxedAgentManager>()?;
+                if let Err(e) = notify_agents(
+                    agent_manager,
+                    hostname.as_str(),
+                    &target_agents.updates,
+                    &target_agents.disables,
+                )
+                .await
+                {
+                    error!("Failed to notify agents for node {i} to be updated. However, the failure does not affect the node application operation.\nDetails:\n{e:?}",);
```
sehkone, line 125: The second sentence doesn't seem to convey exactly what you intended. It sounds like you're saying, "No problem, just ignore this." Could you suggest a better way to avoid misleading readers?

src/graphql/node/control.rs
```diff
@@ -145,20 +168,25 @@ pub struct ApplyNodeResponse {
...
+                async_graphql::Error::new(format!(
+                    "Failed to notify agent for config update {agent_key}: {e}"
+                ))
+            })
     });
 
     // TODO: We need to implement the logic to disable agents. For now, we will only log the
     // message.
-    info!("Agents {disable_agent_keys:?} need to be notified to be disabled, but disabling logic is not yet implemented");
+    info!("Agents {disable_agent_ids:?} need to be notified to be disabled, but disabling logic is not yet implemented");
```
sehkone, line 189: The second sentence doesn’t seem necessary to me. Since log messages are visible to outside partners and even users, I think it's inappropriate to reveal anything that should have been implemented.

src/graphql/node/control.rs
```diff
@@ -99,37 +99,60 @@ impl NodeControlMutation {
...
+                    chrono::Utc::now(),
+                );
+            }
         }
 
         Ok(ApplyNodeResponse { id, giganto_draft })
     }
 }
 
+fn format_agent_updates(node: &Node, target_agents: &NotificationTarget) -> String {
```
sehkone, line 140: Could you explain why the log message includes all the content of the agents' `config`s and `draft`s? Some of these items are usually not visible to users, as the UI only shows what can be modified by them, but they become visible because of this message. If you're considering keeping a history of all the changes, I think it should be stored in the database rather than in log messages.

src/graphql/node/control.rs
```diff
@@ -1296,7 +1324,114 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_apply_node_disables_agents() {}
+    async fn test_apply_node_emtpy_hostname() {
```
sehkone, line 1327: Typo

src/graphql/node/control.rs
sophie-cluml: This log was created to follow this guide

> [Central server] logs the apply actions along with settings config, settings_draft draft, and the timestamp.

src/graphql/node/control.rs
sophie-cluml: It is treated the same as we treat the failures of broadcast customer change and agent notification, which does not return error but only leave a log.

src/graphql/node/control.rs
sophie-cluml: Done

src/graphql/node/control.rs
sophie-cluml: I made a change. Could you check please?

src/graphql/node/control.rs
sophie-cluml: Done

src/graphql/node/control.rs
sophie-cluml: Thank you, done

src/graphql/node/control.rs
sehkone: It looks like I wrote the sentence. I apologize for the confusion caused by using the term "log" too loosely. When I used "log," I was thinking of its literal meaning—saving information—rather than in the syslog style. I intended that the specific method of storing could be decided based on the context. In this case, I don’t think the `config`s and `draft`s should be written to the log. As for how to handle them, I believe we can decide that later.

src/graphql/node/control.rs
sehkone: Got it. I forgot that we decided to do it this way.

src/graphql/node/control.rs
sophie-cluml: Logs on `config` and `draft` are removed.

sehkone: > Logs on `config` and `draft` are removed.

I still see them.

sophie-cluml: I am sorry for the mistake. 🙇 It's done from my side now.

sehkone (approved):

PR #312, opened 2024-10-01
Introduce a new GraphQL API, `ipLocationList`

sehkone:
The UI will add a dashboard feature where a rotatable and zoomable earth globe displays threat flows, each of which starts from a source IP location and ends at a destination IP location. To enable this feature, the UI needs to retrieve the list of multiple IP locations in a single request, but the current interface only supports retrieving the location for one IP address at a time.

To prevent performance degradation, this API limits the maximum number of IP addresses to 200 for returning location information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- Added the `ipLocationList` GraphQL API for multiple addresses.
+
 ### Changed
 
 - Updated review-database to 0.31.0.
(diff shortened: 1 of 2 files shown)


src/graphql/ip_location.rs
```diff
@@ -37,6 +38,46 @@ impl IpLocationQuery {
...
+                                r.ok().map(|location| IpLocationItem {
+                                    address: address.clone(),
+                                    location,
+                                })
+                            })
+                    })
+                })
+                .collect()
+        } else {
+            Vec::new()
```
msk, line 76: If `Mutex::lock` fails, an empty vector is returned. However, according to the [docs](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock), `Mutex::lock` only fails if another thread panicked while holding the mutex. In this case, I think it would be better to return an error instead of an empty vector, as it's not a normal or expected result.

msk (changes requested):

CHANGELOG.md
```diff
@@ -179,8 +183,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   - `get_resource_usage`
   - `ping`
   - `reboot`
-  The implementor of `AgentManager` is now responsible for providing the
-  behavior for these methods.
+    The implementor of `AgentManager` is now responsible for providing the
```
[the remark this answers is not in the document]
sehkone: The change was automatically made by the "Prettier" formatter, so I rephrased the paragraph.

src/graphql/ip_location.rs
sehkone: Thank you for your suggestion. Done.

msk (approved):

Issue #308, opened 2024-09-23
Change `nodeStatusList` GraphQL API to include fields related to `Node`


sophie-cluml:
Currently, the `nodeStatusList` GraphQL API retrieves information about nodes from the database but only responds with `id`, `name`, and the associated agent information to the client, regarding `Node`. However, the client relies on the `nodeStatusList` API response to determine whether it is an appropriate time to call `applyNode`. For the client to make a good choice, the client needs to compare the field values of `Node`, such as if `name` matches `name_draft` and if `profile` matches `profile_draft`. Since the `nodeStatusList` API does not currently provide this information, the client lacks the necessary details for the decision. And I think we need to return the info to the client.

PR #321, opened 2024-10-04
Update `nodeStatusList` GraphQL API

sophie-cluml:
Closes: #308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Updated review-database to 0.31.0.
+- Updated `nodeStatusList` GraphQL API response to include `nameDraft`,
+  `profile`, `profileDraft`, and `gigantoDraft`, offering a more comprehensive
+  view of the node’s status.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Updated review-database to 0.31.0.
+- Updated `nodeStatusList` GraphQL API response to include `nameDraft`,
+ `profile`, `profileDraft`, and `gigantoDraft`, offering a more comprehensive
+ view of the node’s status.
```
sehkone, line 19: This alignment doesn't look good. I suspect you didn't install Prettier.

sophie-cluml: Thank you for the check, and I found that we might need to add this:
image

I had been using Prettier for about a month, but it seems like it did not work as I expected. However, with the addition of this setting, I was able to get the desired result.

sehkone (approved):

CHANGELOG.md
sehkone: I guess you turned off Prettier manually by clicking the icon at the right bottom corner. You can turn it on by clicking it again or setting it in the file as you did.

PR #324, opened 2024-10-07
Make the GraphQL APIs check the pagination params in the same manner

### LINKED ISSUE #323: GraphQL queries `savedOutliers` and `rankedOutliers` panic with negative pagination parameters
The GraphQL queries `savedOutliers` and `rankedOutliers` are causing a panic when a web client provides negative values for the `first` or `last` arguments. This is a critical issue that needs to be addressed to prevent unexpected server termination.

## Current Behavior

- When negative values are provided for `first` or `last` arguments in `savedOutliers` and `rankedOutliers` queries, the server panics.
- The panic occurs in the type conversion from `i32` to `usize`.

## Expected Behavior

- The queries should handle negative values gracefully, similar to other pagination functions in the crate.
- Proper error handling should be implemented instead of panicking.

## Detailed Analysis

1. `savedOutliers` In `graphql/outliers.rs`:
   - `OutlierQuery::saved_outliers` calls `load_outliers`, passing `first` and `last` as `Option<i32>`.
   - `load_outliers` attempts to convert these to `Option<usize>` using:
     ```rust
     first.map(|value| usize::try_from(value).expect("valid size")),
     last.map(|value| usize::try_from(value).expect("valid size")),
     ```
   - When `value` is negative, `usize::try_from` returns `Err`, causing a panic.

2. Similarly, `rankedOutliers` in `graphql/outliers.rs`:
   - `OutlierQuery::ranked_outliers` calls `load_ranked_outliers_with_filter`, which panics in:
     ```rust
     first.to_usize().expect("valid size"),
     ```
     and
     ```rust
     last.to_usize().expect("valid size"),
     ```

3. Other queries (e.g., `outliers`) in the same file don't have this issue because they use [`async_graphql::connection::query`](https://docs.rs/async-graphql/latest/async_graphql/types/connection/fn.query.html), which properly handles the conversion from `Option<i32>` to `Option<usize>`.

## Proposed Solution

- Modify `saved_outliers` and `ranked_outliers` to use `async_graphql::connection::query` for parameter handling, consistent with other pagination functions.

## Steps to Reproduce

1. Run review-web's GraphQL playground.
2. Execute a `savedOutliers` or `rankedOutliers` query with a negative value for `first` or `last`.
3. Observe the server panic.

## Environment

- review-web 0.22.0

sehkone:
Closes #323 

In addition to #323, I refactored the code to eliminate repetition and renamed `load_nodes` in graphql.rs to `load_edges_interim` for clarity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,16 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   instead of its serialized form. This change decouples review-web from
   dictating the serialized form of `HostNetworkGroup`, which should be handled
   by the review-protocol crate.
+- Instead of `async_graphql::connection::query`, its wrapper `graphql::query` is
+  now used for all GraphQL APIs to support pagination.
+  - The GraphQL APIs `clusters`, `eventList`, and `models` were changed to no
+    longer accept invalid pagination parameters, such as providing both `after`
+    and `before`, by using `graphql::query`.
+
+### Fixed
+
+- Fixed `savedOutliers` and `rankedOutliers` to properly validate pagination
+  parameters.
 
 ## [0.22.0] - 2024-10-04
 
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -34,6 +34,7 @@ mod triage;
 mod trusted_domain;
 mod trusted_user_agent;
 
+use std::future::Future;
 #[cfg(test)]
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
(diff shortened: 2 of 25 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed `savedOutliers` and `rankedOutliers` to properly validate pagination
+  parameters, and updated all other GraphQL APIs that provide pages to do so in
+  the same manner."
```
msk, line 14:
```suggestion
  the same manner.
```

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
msk, line 8: There are some user-visible changes that aren't documented in the CHANGELOG. Specifically, the APIs that previously allowed pagination within a range between `after` and `before`, e.g., `clusters`, `eventList`, and `models`, now only accept one of these parameters. I'm concerned that this might be a regression, but if dropping this feature was intentional, it would be great to document it in the CHANGELOG, since it affects web clients. Could you please add a note about this change?

msk (changes requested):

CHANGELOG.md
sehkone: As of now, it looks good enough that all APIs just provide this restricted combination of parameters, and it helps maintaining consistency. So, I added a note.

msk (approved):

Issue #293, opened 2024-08-30
Validate request to apply new configuration to a node


sehkone:
The GraphQL resolver `applyNode` receives a node ID and attempts to update the node's information and agent configurations. However, if the node is modified by another administrator in the meantime, the user of the resolver might encounter unexpected results. Therefore, the resolver should receive all the node's information as an argument and reject the request if it doesn't match the data stored in the database. This way, we can ensure that users of `applyNode` don't experience unexpected updates to the node.

I think it would be great to enable this feature after the current node tasks are completed.

sehkone:
This implies that the user already knows all the information about a node including Giganto. Therefore, the resolver will not need to return the draft of Giganto. Instead, it should return whether the user can send a request to Giganto.

PR #326, opened 2024-10-08
Validate node input in `applyNode` GraphQL API

sophie-cluml:
close: #293


src/graphql/node/control.rs
```diff
@@ -869,23 +899,48 @@ mod tests {
         let res = schema
```
sehkone, line 899: This comment should be modified according to the change; No longer returning of gigantoDraft.

sophie-cluml: Thank you so much 🙇 I updated the comment for `apply_node` function accordingly as well.

sehkone (approved):

PR #329, opened 2024-10-14
Refactor code to avoid redundant query validation

### LINKED ISSUE #328: Refactor `validate_pagination_params` to avoid redundancy with async-graphql's internal checks
The `validate_pagination_params` function currently (as of 7fccd0a) performs several checks on pagination parameters (first, last, before, and after) that are already handled by the async-graphql crate. Since async-graphql adheres to the [GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm), it ensures that:

- Only one of `first` and `last` is present.
- The values of `first` and `last` are non-negative integers.

As a result, some of the validations in `validate_pagination_params` are redundant. These checks should be refactored to focus on enforcing additional requirements specific to review-web's API, rather than duplicating checks that async-graphql already handles.

sehkone:
Closes #328 

These two wrappers, `graphql::query` and `graphql::query_with_contraints` resolve the issue and revert the regression affecting several APIs, such as `clusters`.

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -216,47 +216,96 @@ where
     R: Future<Output = Result<Connection<String, Node, ConnectionFields, EmptyFields, Name>, E>>,
     E: Into<async_graphql::Error>,
 {
-    let (first, last) = validate_pagination_params(after.is_some(), before.is_some(), first, last)?;
+    let (first, last) = connection_size(after.is_some(), before.is_some(), first, last)?;
+
     async_graphql::connection::query(after, before, first, last, |after, before, first, last| {
         f(after, before, first, last)
     })
     .await
 }
 
-const DEFAULT_CONNECTION_SIZE: i32 = 100;
+async fn query_with_constraints<Node, ConnectionFields, Name, F, R, E>(
+    after: Option<String>,
+    before: Option<String>,
+    first: Option<i32>,
+    last: Option<i32>,
+    f: F,
+) -> Result<Connection<String, Node, ConnectionFields, EmptyFields, Name>>
+where
+    Node: OutputType,
+    ConnectionFields: ObjectType,
+    Name: ConnectionNameType,
+    F: FnOnce(Option<String>, Option<String>, Option<usize>, Option<usize>) -> R,
+    R: Future<Output = Result<Connection<String, Node, ConnectionFields, EmptyFields, Name>, E>>,
+    E: Into<async_graphql::Error>,
+{
+    extra_validate_pagination_params(
+        after.is_some(),
+        before.is_some(),
+        first.is_some(),
+        last.is_some(),
+    )?;
+    let (first, last) = connection_size(after.is_some(), before.is_some(), first, last)?;
 
-fn validate_pagination_params(
+    async_graphql::connection::query(after, before, first, last, |after, before, first, last| {
+        f(after, before, first, last)
+    })
+    .await
+}
+
+#[derive(Debug, thiserror::Error)]
+enum Error {
+    #[error("Invalid cursor")]
+    InvalidCursor,
+    #[error("The value of first and last must be within 0-100")]
+    InvalidLimitValue,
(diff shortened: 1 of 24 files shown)


src/graphql.rs
```diff
@@ -225,38 +247,33 @@ where
 
 const DEFAULT_CONNECTION_SIZE: i32 = 100;
 
-fn validate_pagination_params(
+fn default_connection_size(
```
msk, line 250: Given the name of this function, shouldn't it guarantee that one of the return values is `Some`? In the current implementation, it may be `(None, None)` if both `after` and `before` is `true`.

sehkone: I discovered that `clusters` and a few more APIs use `slicing::limit` to validate that both `first` and `last` cannot be used simultaneously (which is redundant) as well as to limit the connection size. Initially, I didn't consider limiting the size to avoid performance degradation. Based on what I've learned, I refactored the code to follow these criteria:

- If the connection size is not specified, provide a default size, except the case when `after` and `before` are both given. 
- If the connection size is specified but exceeds the limit, notify the users.
- Every API should use either `graphql::query` or `graphql::query_with_constraints`. The former relies on only `async_graphql::query` for pagination parameter validation, while the latter adds more constraints. Both ensure the above two points by calling `graphql::connection_size`

src/graphql.rs
```diff
@@ -216,47 +216,100 @@ where
...
-    }
-    if let Some(last) = last {
-        if last < 0 {
-            return Err("The \"last\" parameter must be a non-negative number".into());
-        }
+    first: Option<i32>,
+    last: Option<i32>,
+) -> Result<(Option<i32>, Option<i32>), Error> {
+    match (after, before, first, last) {
+        (true, true, None, None) => Err(Error::MissingPaginationBoundaries),
```
msk, line 284: Is this error correct? It seems counter-intuitive to call it "**missing** pagination boundaries" when both `after` and `before` are `true`.

sehkone: I replaced it with `MissingPaginationSize`.

msk: The error message for `MissingPaginationSize` (“You must provide a first or last value to properly paginate a connection.”) implies that either first or last must always be provided. However, in the second and third `match` arms, neither `first` nor `last` is provided, and the function still accepts this by using a default size. This seems inconsistent with the error message. Either the error message should be updated to reflect the specific conditions that lead to the error (e.g., when both after and before are true), or the logic should enforce that one of first or last is always required, as the current error message suggests. Could you clarify this?

sehkone: I agree with your point. Based on your feedback, I adjusted how the connection size is handled:

- When users do not specify a connection size, it is assumed that they want to retrieve as many items as the server can provide. Therefore, `MAX_CONNECTION_SIZE` is used, and the previous `DEFAULT_CONNECTION_SIZE` has been removed.
- If both `after` and `before` are specified, users are likely not specifying a size because they want all items between the two cursors. To prevent performance degradation, `MAX_CONNECTION_SIZE` should be applied.
- As a result, `MissingPaginationSize` has been naturally removed.

What do you think?

src/graphql.rs
```diff
@@ -216,47 +216,96 @@ where
...
-        (_, true, _, true) => return Err("cannot provide both `last` and `after`".into()),
-        (false, false, false, _) => {
-            first = Some(DEFAULT_CONNECTION_SIZE);
-        }
-        (false, false, true, false) => {
-            last = Some(DEFAULT_CONNECTION_SIZE);
-        }
-        _ => {}
+fn limit(len: i32) -> Result<i32, Error> {
+    if (1..=MAX_CONNECTION_SIZE).contains(&len) {
```
msk, line 289: According to the GraphQL Cursor Connections Specification, `first` and `last` can also be zero. Would it make sense to update the range to `0..=MAX_CONNECTION_SIZE` to accommodate this?

sehkone: I didn't know that, and fixed it.

msk (approved):

PR #344, opened 2024-10-27
Use ring as crypto provider in rustls for consistency

### LINKED ISSUE #343: Ubuntu 20.04 does not support the recent axum-server
The recent review-web updated axum-server from version 0.6 to 0.7. However, in Ubuntu 20.04, the software cannot be compiled due to its lack of support for aws-lc-sys v0.22.0, which axum-server now requires. This issue does not occur in Ubuntu 22.04.

Since we need to support Ubuntu 20.04 until our clients upgrade their OS, I think it would be best to specify different versions of axum-server based on Linux versions through features in Cargo.toml. Alternatively, if there’s a way to detect Linux versions, that would be preferable, though I’m unsure if such a method exists.

@msk, could you share any thoughts on this?

msk:
The crypto provider used by rustls in review-web has been changed to ring, which is already used by review-database. This avoids compiling multiple crypto providers and promotes consistency across dependencies.

Resolves #343.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ async-graphql = { version = "7", features = ["chrono", "string_number"] }
 async-graphql-axum = "7"
 async-trait = "0.1"
 axum = { version = "0.7", features = ["macros", "tokio", "ws"] }
-axum-server = { version = "0.7", features = ["tls-rustls"] }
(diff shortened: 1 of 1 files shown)


sehkone:
How simple it is! Thank you so much.

sehkone (approved):

Issue #346, opened 2024-10-30
Make `SamplingPolicy` expose `node` in GraphQL API response in `samplingPolicy` and `samplingPolicyList` GraphQL APIs


sophie-cluml:
The function to expose `node` is currently missing in the `samplingPolicy` and `samplingPolicyList` GraphQL API responses, as seen in 

https://github.com/aicers/review-web/blob/d7a30a3f16144936d94f9209838f130f46a5b5f0/src/graphql/sampling.rs#L132-L177 

This omission is causing errors, so this issue aims to address and resolve the missing `node` field in the `SamplingPolicy` response.

PR #347, opened 2024-10-30
Fix sampling policy related APIs

sophie-cluml:
Close: #346 

Modified `samplingPolicy` and `samplingPolicyList` GraphQL API to have `node` field in their response.

### DIFF
--- a/src/graphql/sampling.rs
+++ b/src/graphql/sampling.rs
@@ -163,6 +163,10 @@ impl SamplingPolicy {
         self.inner.dst_ip.as_ref().map(ToString::to_string)
     }
 
+    async fn node(&self) -> Option<String> {
+        self.inner.node.clone()
+    }
+
     async fn column(&self) -> Option<StringNumber<u32>> {
         self.inner.column.map(StringNumber)
     }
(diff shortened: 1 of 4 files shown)


src/graphql/sampling.rs
```diff
@@ -163,6 +163,10 @@ impl SamplingPolicy {
         self.inner.dst_ip.as_ref().map(ToString::to_string)
     }
 
+    async fn node(&self) -> Option<String> {
+        self.inner.node.as_ref().map(ToString::to_string)
```
sehkone, line 167: Could you explain why you didn't use `clone`?

sophie-cluml: Thank you for the comment. I changed it to clone, which I agree as simpler and straightforward.

src/graphql/sampling.rs
```diff
@@ -458,6 +465,7 @@ mod tests {
                         interval: FIFTEEN_MINUTES,
                         period: ONE_DAY,
                         offset: 0,
+                        node: "collector",
```
sehkone, line 468: The word "collector", as I mentioned elsewhere, doesn't quite capture the meaning we intended. We've already decided to go with a much better name: "sensor".

sophie-cluml: Done. I replaced "collector" with "sensor" in group.rs and event.rs.

sehkone (approved):

PR #364, opened 2024-12-02
Update GraphQL cursor to use `OpaqueCursor`

### LINKED ISSUE #355: Simplify cursor handling using async-graphql's `OpaqueCursor`
review-web manually encodes and decodes cursors using `graphql::encode_cursor` and `graphql::decode_cursor`. However, async-graphql provides a convenient [`OpaqueCursor`](https://docs.rs/async-graphql/latest/async_graphql/types/connection/struct.OpaqueCursor.html) type that handles base64 encoding and decoding.

It would be better to simplify the codebase by using `OpaqueCursor` instead of manual encoding and decoding. This would not only reduce the amount of code but also make it more maintainable and less prone to errors.

henry0715-dev:
Close #355

The following sources are the main areas that have been modified.
graphql.rs
graphql directory sub-source

The rest of the changes are fixes for Clippy errors caused by the
Rust version update.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -48,7 +48,7 @@ impl Config {
 
     pub(crate) fn configure_reverse_proxies(
         store: &Arc<RwLock<Store>>,
-        client: &Option<reqwest::Client>,
+        client: Option<&reqwest::Client>,
         reverse_proxies: &[Self],
     ) -> Vec<(ArchiveState, Router<ArchiveState>)> {
         reverse_proxies
@@ -57,10 +57,10 @@ impl Config {
                 (
                     crate::archive::ArchiveState {
                         store: store.clone(),
-                        client: client.clone(),
+                        client: client.cloned(),
                         config: rp.clone(),
                     },
-                    crate::archive::reverse_proxy(store.clone(), client.clone(), rp.clone()),
+                    crate::archive::reverse_proxy(store.clone(), client.cloned(), rp.clone()),
                 )
             })
             .collect()
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -40,13 +40,12 @@ use std::future::Future;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
 
-use async_graphql::connection::{ConnectionNameType, CursorType, EdgeNameType};
+use async_graphql::connection::{ConnectionNameType, CursorType, EdgeNameType, OpaqueCursor};
 use async_graphql::{
     connection::{Connection, Edge, EmptyFields},
     Context, Guard, MergedObject, MergedSubscription, ObjectType, OutputType, Result,
 };
 use chrono::TimeDelta;
-use data_encoding::BASE64;
 use num_traits::ToPrimitive;
 #[cfg(test)]
 use review_database::HostNetworkGroup;
@@ -236,13 +235,23 @@ async fn query_with_constraints<Node, ConnectionFields, Name, F, R, E>(
     first: Option<i32>,
     last: Option<i32>,
     f: F,
-) -> Result<Connection<String, Node, ConnectionFields, EmptyFields, Name>>
+) -> Result<Connection<OpaqueCursor<Vec<u8>>, Node, ConnectionFields, EmptyFields, Name>>
 where
     Node: OutputType,
     ConnectionFields: ObjectType,
     Name: ConnectionNameType,
-    F: FnOnce(Option<String>, Option<String>, Option<usize>, Option<usize>) -> R,
-    R: Future<Output = Result<Connection<String, Node, ConnectionFields, EmptyFields, Name>, E>>,
+    F: FnOnce(
+        Option<OpaqueCursor<Vec<u8>>>,
+        Option<OpaqueCursor<Vec<u8>>>,
+        Option<usize>,
+        Option<usize>,
+    ) -> R,
+    R: Future<
+        Output = Result<
+            Connection<OpaqueCursor<Vec<u8>>, Node, ConnectionFields, EmptyFields, Name>,
+            E,
+        >,
+    >,
     E: Into<async_graphql::Error>,
 {
     extra_validate_pagination_params(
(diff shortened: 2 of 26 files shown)


src/graphql/node/status.rs
```diff
@@ -67,9 +68,14 @@ async fn load(
         let (resource_usage, ping) =
             fetch_resource_usage_and_ping(agent_manager, hostname, is_manager).await;
 
+        let encoded;
+        {
+            let key = node.unique_key();
+            encoded = OpaqueCursor(key.to_vec()).encode_cursor();
+        }
```
sehkone, line 75: Why would you just declare `encoded` in advance and initialize it later? Additionally, I don't see any advantage to limit the lifetime of `key` within the block.

src/archive.rs
```diff
@@ -46,9 +46,10 @@ impl Config {
         &self.uri
     }
 
+    #[allow(clippy::ref_option, clippy::clone_on_copy)]
```
msk, line 49: This `allow` attribute is no longer needed, since you've already fixed the clippy warnings.

msk (changes requested):
@henry0715-dev, thanks for taking on this issue! This PR is definitely a step in the right direction, and I appreciate that you've replaced our manual encoding/decoding functions with `OpaqueCursor`'s routines. I wouldn't mind seeing this PR merged (after addressing the code-level comments, of course).

However, I wanted to point out that the eventual goal is to take it a step further and completely eliminate the need for manual encoding/decoding calls. To demonstrate what I mean, I've created a branch that shows how to integrate `OpaqueCursor` more deeply into the GraphQL API.

For example, take a look at how `ClusterQuery::clusters` is implemented in [my branch](https://github.com/msk/review-web/blob/replace-indexed-key-with-opaque-cursor/src/graphql/cluster.rs). Notice how the `query` function now calls `load` with decoded `after` and `before` values (which are no longer `Option<String>`), and returns the cursor before encoding (in this case, `OpaqueCursor<(i32, i64)>`).

By doing it this way, we can simplify the code even further and make it more maintainable. Just something to keep in mind for future improvements. Again, thanks for your work on this PR.

src/archive.rs
henry0715-dev: I'll remove it.
Thank you.

src/graphql/node/status.rs
henry0715-dev: I'll change it.
Thank you.

henry0715-dev:
> @henry0715-dev, thanks for taking on this issue! This PR is definitely a step in the right direction, and I appreciate that you've replaced our manual encoding/decoding functions with `OpaqueCursor`'s routines. I wouldn't mind seeing this PR merged (after addressing the code-level comments, of course).
> 
> However, I wanted to point out that the eventual goal is to take it a step further and completely eliminate the need for manual encoding/decoding calls. To demonstrate what I mean, I've created a branch that shows how to integrate `OpaqueCursor` more deeply into the GraphQL API.
> 
> For example, take a look at how `ClusterQuery::clusters` is implemented in [my branch](https://github.com/msk/review-web/blob/replace-indexed-key-with-opaque-cursor/src/graphql/cluster.rs). Notice how the `query` function now calls `load` with decoded `after` and `before` values (which are no longer `Option<String>`), and returns the cursor before encoding (in this case, `OpaqueCursor<(i32, i64)>`).
> 
> By doing it this way, we can simplify the code even further and make it more maintainable. Just something to keep in mind for future improvements. Again, thanks for your work on this PR.

Thank you for the detailed explanation.
We have made changes to suit your requirements.
I would appreciate it if you could check if the revised PR is consistent with your suggested direction.

src/graphql.rs
```diff
@@ -432,8 +420,13 @@ where
...
     connection.edges.extend(nodes.into_iter().map(|node| {
         let Ok(node) = node else { unreachable!() };
-        let encoded = encode_cursor(node.unique_key().as_ref());
-        Edge::new(encoded, node.into())
+        let cursor;
+        {
+            let key = node.unique_key();
+            let key = key.as_ref().to_vec();
+            cursor = OpaqueCursor(key);
+        }
```
sehkone, line 428: This part has the same issue that I mentioned earlier. Could you correct it?

henry0715-dev: An error occurs when correcting the source as follows.
```
        let key = node.unique_key();
        let key = key.as_ref().to_vec();
        Edge::new(OpaqueCursor(key), node.into())
```
error[E0505]: cannot move out of `node` because it is borrowed
borrow might be used here, when `key` is dropped and runs the destructor for type `<R as UniqueKey>::AsBytes<'_>`

If you have any recommendations, please let me know.

henry0715-dev: I will make the change as below.
```
  let key = node.unique_key().as_ref().to_vec();
  Edge::new(OpaqueCursor(key), node.into())
```

sehkone:
@henry0715-dev,

I guess you haven't set up your editor properly regarding limiting the number of characters in the commit message and body according to our rules on how to write commit messages. Could you check it and correct this commit message?

cc: @sophie-cluml

henry0715-dev:
> @henry0715-dev,
> 
> I guess you haven't set up your editor properly regarding limiting the number of characters in the commit message and body according to our rules on how to write commit messages. Could you check it and correct this commit message?
> 
> cc: @sophie-cluml

You have changed the commit message.
I'll be careful.
Thank you.

sehkone:
>  I'll be careful. 

Thank you for the modification, and I appreciate your comment about being careful. However, this is just one of many trivial conventions we are expected to follow, and I don’t think it’s realistic to rely solely on being careful or cautious to get everything perfect. I strongly suggest that every software engineer establish their own automated routines and processes to avoid missing these small details.

henry0715-dev:
> Thank you for the modification, and I appreciate your comment about being careful. However, this is just one of many trivial conventions we are expected to follow, and I don’t think it’s realistic to rely solely on being careful or cautious to get everything perfect. I strongly suggest that every software engineer establish their own automated routines and processes to avoid missing these small details.

I will apply it to writing a commit message with an automation tool.
Thank you.

PR #367, opened 2024-12-06
Use pass-by-value for SocketAddr

msk:
Since `SocketAddr` is `Copy` and small enough, it's better to pass it by value.


sophie-cluml (approved):

sehkone:
Since this code was created by @henry0715-dev, I would like to share this with him.

Issue #359, opened 2024-11-24
Do not treat conversion failure to `Config` as `None`


sehkone:
https://github.com/aicers/review-web/blob/fee9a082b10d0ebef793efd1452a32ae734423bb/src/graphql/node/input.rs#L147

https://github.com/aicers/review-web/blob/fee9a082b10d0ebef793efd1452a32ae734423bb/src/graphql/node/input.rs#L155

@BLYKIM

As we all know, `None` has a special meaning: it indicates that the corresponding agent will not run on the node. However, the current two lines also treat conversion failures as `None`, even when this is not the intended case. A failure to convert to `Config` signifies a critical error, which should be properly reported, and the procedure should not proceed with `None`.

We should be cautious when using `ok()`, as it essentially ignores errors.

cc: @sophie-cluml

PR #373, opened 2024-12-18
Handle configuration conversion failures explicitly

BLYKIM:
Close: #359

### DIFF
--- a/src/graphql/node/input.rs
+++ b/src/graphql/node/input.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, fmt};
 
 use anyhow::Context as AnyhowContext;
-use async_graphql::{types::ID, InputObject, Result};
+use async_graphql::{types::ID, Error, InputObject, Result};
 
 use super::{AgentKind, AgentStatus};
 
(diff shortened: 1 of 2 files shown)


src/graphql/node/input.rs
```diff
@@ -138,27 +138,50 @@ pub(super) fn create_draft_update(
...
-                    status: agent_draft.status.into(),
-                    config,
-                    draft: agent_draft.draft.and_then(|draft| draft.try_into().ok()),
-                }
+    let agents: Vec<review_database::Agent> =
+        new.agents
+            .map(|new_agents| {
+                new_agents
+                    .into_iter()
+                    .map(|agent_draft| {
```
sophie-cluml, line 146: I think `agent_draft` can bring misunderstanding in this context, because it looks like the draft configuration of the, but it is actually not. I think we can use `new_agent` instead.

sophie-cluml (approved):

sehkone:
@BLYKIM, could you rebase this? Rust has been updated.

sehkone:
@sophie-cluml, @BLYKIM

As I mentioned [here](https://github.com/aicers/review-web/pull/381#discussion_r1912726977), I think we need to include returning specific information about which agents caused the failures. If you agree, could you create an issue for it?

sehkone:
@BLYKIM @sophie-cluml,

Since the commit title exceeds the limit according to our rules, I modified it.

Issue #357, opened 2024-11-20
Remove `unwrap` & Add clippy lint option to check the use of `unwrap`


dayeon5470:
related issue: https://github.com/aicers/patio/issues/211

PR #388, opened 2025-01-14
Refactor `unwrap`

BLYKIM:
Close: #357

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -113,7 +113,10 @@ async fn process_request(
     tail: Option<Path<String>>,
     req: Request<Body>,
 ) -> Result<Response, Error> {
-    let (_store, client, config) = (state.store, state.client.unwrap(), state.config);
+    let client = state
+        .client
+        .ok_or_else(|| Error::ServiceUnavailable("Client was not initialized".to_string()))?;
+    let config = state.config;
 
     let url = match tail {
         Some(Path(tail)) => format!("{}/{tail}", config.uri()),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,11 +116,14 @@ where
             let completed = shutdown_completed.clone();
             tokio::spawn(graceful_shutdown(handle.clone(), wait_shutdown));
             tokio::spawn(async move {
-                axum_server::bind_rustls(config.addr, tls_config)
+                if let Err(e) = axum_server::bind_rustls(config.addr, tls_config)
                     .handle(handle)
                     .serve(router.into_make_service_with_connect_info::<SocketAddr>())
                     .await
-                    .unwrap();
+                {
+                    panic!("Web server encountered an error: {e:?}");
+                };
+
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -116,11 +116,15 @@ where
...
             tokio::spawn(graceful_shutdown(handle.clone(), wait_shutdown));
             tokio::spawn(async move {
-                axum_server::bind_rustls(config.addr, tls_config)
+                match axum_server::bind_rustls(config.addr, tls_config)
                     .handle(handle)
                     .serve(router.into_make_service_with_connect_info::<SocketAddr>())
                     .await
-                    .unwrap();
+                {
+                    Ok(()) => info!("Web server stopped gracefully."),
```
sehkone, line 124: Does this message accurately report the status? It may imply that there is no response from the server at all, as the server has already stopped at this point.

BLYKIM: I have updated the log messages, but I might be mistaken due to my limited understanding of this part of the code. Could you kindly recommend a more appropriate log message if needed? Thank you!

BLYKIM: The current message is logged when the server completes handling all requests and stops accepting new connections. To my understanding, this indicates the server has halted its main operations. However, if there are additional shutdown steps (e.g., resource cleanup), they might not be reflected in this message. I can further investigate if needed.

sophie-cluml: Based on my findings on `axum_server::server::Server::serve`, at the time the `Ok(())` match arm is hit, the GraphQL server is already shut down. However, `axum_server::server::Server::serve` logic says that the GraphQL server might have shut down gracefully or forcefully, depending on the circumstances. So we cannot guarantee that the Ok match arm means a graceful shutdown.

Given this, I think below might be one of the possible approaches:  
- If `axum_server::server::Server::serve` returns `Ok(())`, I think we could choose not to log anything i as no logs were previously logged in this case.  
- If `axum_server::server::Server::serve` returns an error, I believe we can `panic!`. This is because such errors only occur when binding to the address (including the port) fails, making it impossible for `review-web` to function as a GraphQL server.  
  - Alternatively, if we decide not to panic, I believe the documentation comment for `lib::serve` should be updated to reflect this, as it currently states that the function will panic in such situations.

src/archive.rs
```diff
@@ -113,7 +113,10 @@ async fn process_request(
     tail: Option<Path<String>>,
     req: Request<Body>,
 ) -> Result<Response, Error> {
-    let (_store, client, config) = (state.store, state.client.unwrap(), state.config);
+    let client = state
+        .client
+        .ok_or_else(|| Error::ServiceUnavailable("Client was not initialized".to_string()))?;
+    let (_store, config) = (state.store, state.config);
```
sophie-cluml, line 119: It seems like `_store` variable is unused. Could you remove it please?

sophie-cluml (approved):

Issue #389, opened 2025-01-17
Modify to broadcast the correct internal network list for each Semi-supervised Engine


kimhanbeom:
Related issue: discussion repository issue 236

## Background

- Previously, it was possible to register Central Management Server as the target of node registration, and all Semi-supervised Engine received the internal network list corresponding to Central Management Server's customer id.
- However, currently, Central Management Server has been excluded as a target for node registration, and as a result, it is not possible to send a valid internal network list to Semi-supervised Engine. 
- Therefore, if we need to send an internal network to Semi-supervised Engine, we need to make a modification so that we can send an internal network list that matches the customer id of each Semi-supervised Engine.
- The main direction of the modification is as follows.

## Task
- Create a `HashMap<u32, Vec<String>>` with the customer id as the key and the key(agent@hostname) of the agents belonging to the customer id as the value. 
- When the `Update Customer`, `Remove Customer` GraphQL APIs are called, the internal network and the agent keys are obtained via the customer id for that customer.
- Create a `NetworksAgentKeysPair` consisting of information(agent keys) about the target(Semi-supervised Engine) that will receive the network and the internal network information that will be sent to the target.
- Rename the `broadcast_internal_networks` method of to `AgentManager` trait to `send_agent_specific_internal_networks` as the functionality of `broadcast_internal_networks` changes from broadcast to fine-targeting nodes and agents using agent and hostnames to send.
- Change the argument type of the `send_agent_specific_internal_networks` method from `HostNetworkGroup` to `NetworksTargetAgentKeysPair` array. This change will allow the Central Management Server that implements `send_agent_specific_internal_networks` to provide the internal networks corresponding to the agent information of the Semi-supervised Engine.

PR #391, opened 2025-01-20
Modify to broadcast the correct internal networks

kimhanbeom:
- Add `NetworksAgentKeysPair` structure that stores the internal
  networks and a list of agent information that use that networks.
- Rename the `broadcast_internal_networks` method of to `AgentManager`
   trait to `send_agent_specific_internal_networks`.
- Change the argument type of `send_agent_specific_internal_networks`
  method from `HostNetworkGroup` to `NetworksTargetAgentKeysPair` array.
- Remove the part that broadcasts when registering a Central Management
  Server node from the `insertNode` GraphQL API.
- Modify the `removeCustomers`, `updateCustomer` GraphQL API to
  generate information about internal networks as
  `Vec<NetworksAgentKeysPair>` instead of `HostNetworkGroup`.
- Rename `get_customer_id_of_node` to `agent_keys_by_customer_id` as
  the functionality of `get_customer_id_of_node` has changed.

Close: #389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,20 @@ Versioning](https://semver.org/spec/v2.0.0.html).
   from `&[u8]` to `&[String]`. Implementors of `AgentManager` will need to
   update their implementations. This change simplifies the API by removing
   serialization concerns from callers.
+- Modified to broadcast the correct internal network list for each
+  Semi-supervised Engine. The changes are as follows.
+  - Renamed the `broadcast_internal_networks` method of to `AgentManager` trait
+    to `send_agent_specific_internal_networks` as the functionality of
+    `broadcast_internal_networks` changes from broadcast to fine-targeting nodes
+    and agents using agent keys and hostnames to send.
+  - Changed the argument type of the `send_agent_specific_internal_networks`
+    method from `HostNetworkGroup` to `NetworksTargetAgentKeysPair` array. This
+    change will allow the Central Management Server that implements
+    `send_agent_specific_internal_networks` to provide the internal networks
+    corresponding to the agent information of the Semi-supervised Engine.
+  - Renamed `get_customer_id_of_node` to `agent_keys_by_customer_id` as the
+    functionality of `get_customer_id_of_node` has changed. The function returns
+    agent info list by all customer id.
 
 ### Removed
 
--- a/src/graphql/customer.rs
+++ b/src/graphql/customer.rs
@@ -9,10 +9,45 @@ use async_graphql::{
 };
 use chrono::{DateTime, Utc};
 use review_database::{self as database, Store};
+use serde::{Deserialize, Serialize};
 use tracing::error;
 
-use super::{get_customer_id_of_node, BoxedAgentManager, Role, RoleGuard};
+use super::node::SEMI_SUPERVISED_AGENT;
+use super::{agent_keys_by_customer_id, BoxedAgentManager, Role, RoleGuard};
 use crate::graphql::query_with_constraints;
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct NetworksTargetAgentKeysPair {
+    networks: database::HostNetworkGroup,
+    target_agent_keys: Vec<String>,
+}
+
+impl NetworksTargetAgentKeysPair {
+    #[must_use]
+    pub fn new(
+        networks: database::HostNetworkGroup,
+        agent_keys: Vec<String>,
+        target_agent: &str,
+    ) -> Self {
+        Self {
+            networks,
+            target_agent_keys: agent_keys
+                .into_iter()
+                .filter(|s| s.starts_with(target_agent))
+                .collect(),
+        }
+    }
+
+    #[must_use]
+    pub fn target_agent_keys(&self) -> &[String] {
+        &self.target_agent_keys
+    }
+
+    #[must_use]
+    pub fn networks(&self) -> &database::HostNetworkGroup {
+        &self.networks
+    }
+}
 
 #[derive(Default)]
 pub(super) struct CustomerQuery;
@@ -101,13 +136,13 @@ impl CustomerMutation {
         ctx: &Context<'_>,
         #[graphql(validator(min_items = 1))] ids: Vec<ID>,
     ) -> Result<Vec<String>> {
-        let (registered_customer_is_removed, removed) = {
+        let (removed_customer_networks, removed) = {
             let store = crate::graphql::get_store(ctx).await?;
             let map = store.customer_map();
             let network_map = store.network_map();
 
-            let customer_id = get_customer_id_of_node(&store).ok().flatten();
-            let mut registered_customer_is_removed = false;
+            let customer_id_hash = agent_keys_by_customer_id(&store)?;
+            let mut removed_customer_networks = Vec::new();
             let mut removed = Vec::<String>::with_capacity(ids.len());
             for id in ids {
                 let i = id.as_str().parse::<u32>().map_err(|_| "invalid ID")?;
@@ -120,20 +155,22 @@ impl CustomerMutation {
                 };
                 removed.push(name);
 
-                if customer_id == Some(i) {
-                    registered_customer_is_removed = true;
+                if let Some(agent_keys) = customer_id_hash.get(&i) {
+                    let network_list = NetworksTargetAgentKeysPair::new(
+                        database::HostNetworkGroup::new(vec![], vec![], vec![]),
+                        agent_keys.clone(),
+                        SEMI_SUPERVISED_AGENT,
+                    );
+                    removed_customer_networks.push(network_list);
                 }
             }
 
-            (registered_customer_is_removed, removed)
+            (removed_customer_networks, removed)
         };
 
-        if registered_customer_is_removed {
-            if let Err(e) = broadcast_customer_networks(
-                ctx,
-                &database::HostNetworkGroup::new(vec![], vec![], vec![]),
-            )
-            .await
+        if !removed_customer_networks.is_empty() {
+            if let Err(e) =
+                send_agent_specific_customer_networks(ctx, &removed_customer_networks).await
             {
                 error!("failed to broadcast internal networks. {e:?}");
             }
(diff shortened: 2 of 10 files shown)


kimhanbeom:
Change it to WIP because it still needs work.

kimhanbeom:
The modifications are complete, and I've removed the WIP.

src/graphql/node/control.rs
```diff
@@ -234,11 +236,21 @@ async fn update_db(
...
+            .agents
+            .iter()
+            .filter_map(|agent| gen_agent_info(agent.kind, hostname).ok())
+            .collect::<Vec<String>>();
         let Ok(customer_id) = customer_id.parse::<u32>() else {
             error!("Failed to parse customer ID from node {i} for broadcasting customer change");
             return;
         };
-        if let Err(e) = broadcast_customer_change(ctx, customer_id).await {
+        if let Err(e) = broadcast_customer_change(ctx, customer_id, agent_list).await {
```
sophie-cluml, line 253: - I don't think it is necessary to include agents that are not Semi-supervised Engine in the `agent_list`. What do you think about minimizing the data transfer here?
- I think we are moving out of the idea of broadcasting, with this PR. In the past, we broadcasted the customer change to all Semi-supervised Engines. But, with this PR, we fine-target the node and the agent with the agent key and the hostname. So I think we should consider updating the name, to align with our idea shift.

src/graphql/node/control.rs
```diff
@@ -234,11 +236,21 @@ async fn update_db(
 
 async fn broadcast_customer_change_if_needed(ctx: &Context<'_>, i: u32, node: &NodeInput) {
     if let Some(customer_id) = customer_id_to_broadcast(node) {
+        let hostname = node
+            .profile_draft
+            .as_ref()
+            .expect("When customer_id exists, nodeInput.profile_draft means Some, which means that the values of the other fields in the NodeProfileInput also exist. Therefore, their values are always valid.")
```
sophie-cluml, line 242: I think backticks will be beneficial here. Could you update please?

src/graphql/customer.rs
```diff
@@ -9,11 +9,18 @@ use async_graphql::{
...
 use tracing::error;
 
-use super::{get_customer_id_of_node, BoxedAgentManager, Role, RoleGuard};
+use super::{get_agent_info_by_customer_id, BoxedAgentManager, Role, RoleGuard};
 use crate::graphql::query_with_constraints;
 
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct NetworksByAgentList {
+    pub networks: database::HostNetworkGroup,
+    pub agent_list: Vec<String>,
```
sophie-cluml, line 21: - I would like to suggest using `agent_keys` as the name, to maintain consistent naming in the `AgentManager`. I assume the Strings in the vector are in the format of {agent_kind}@{hostname}, which for example would be "hog@localhost". and the `AgentManager::update_config`'s `_agent_key` has that format.
- Also, what do you think about taking `agent_list` out of the struct and having it as a function parameter of `broadcast_internal_networks`? By doing so, I think it would be more obvious that we need to rename the `broadcast_internal_networks` to something else, while maintaining the consistent function parameter levels in the `AgentManager` trait.

sehkone: Regardless of @sophie-cluml's suggestion, I think it would be better to rename `NetworkByAgentList` to something else, because it implies the existence of networks for each agent list, which could be interpreted as a primary key.

sehkone: Meanwhile, I think `Vec<NetworksByAgentList>` doesn't seem useful because it's not searchable. Neither `networks` nor `agent_list` could serve as a primary key. In that regard, I still believe the pair of `customer_id` and `agent_list` would be more meaningful from a data structure perspective.

sehkone: Does the method `broadcast_customer_networks` need the entire list of `NetworksByAgentList`, or can it work with a specific `networks` instead?

kimhanbeom: After talking to @sehkone about the `NetworksByAgentList` structure, we decided to just rename the structure.  
- `NetworksByAgentList`  -> `NetworksAgentKeysPair`

src/graphql/node/control.rs
kimhanbeom: done

src/graphql/node/control.rs
kimhanbeom: done

src/graphql/customer.rs
kimhanbeom: done

CHANGELOG.md
```diff
@@ -57,6 +57,20 @@ Versioning](https://semver.org/spec/v2.0.0.html).
...
+- Modified to broadcast the correct internal network list for each
+  Semi-supervised Engine. The changes are as follows.
+  - Changed the argument type of the `broadcast_internal_networks` method from
+    `HostNetworkGroup` to `NetworksAgentKeysPair` array. This change will
+    allow the Central Management Server that implements
+    `broadcast_internal_networks` to provide the internal networks corresponding
+    to the agent information of the Semi-supervised Engine.
+  - Renamed the `broadcast_internal_networks` method of to `AgentManager` trait
+    to `send_agent_specific_internal_networks` as the functionality of
+    `broadcast_internal_networks` changes from broadcast to fine-targeting nodes
```
sophie-cluml, line 69: I think it would be more reader friendly to put this renaming as the first item in the indented list, so that we can use the new name consistently in the following items in the list, given that users usually read from top to bottom. It is because we want to tell the readers that the new `send_agent_specific_internal_networks` accepts the param in `NetworksAgentKeysPair` type.

src/graphql/customer.rs
```diff
@@ -9,11 +9,35 @@ use async_graphql::{
...
 
-use super::{get_customer_id_of_node, BoxedAgentManager, Role, RoleGuard};
+use super::node::SEMISUPERVISED_AGENT;
+use super::{get_agent_key_by_customer_id, BoxedAgentManager, Role, RoleGuard};
 use crate::graphql::query_with_constraints;
 
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct NetworksAgentKeysPair {
+    pub networks: database::HostNetworkGroup,
+    pub agent_keys: Vec<String>,
```
sophie-cluml, line 22: I have one question remaining. It seems like this vector of agent keys will always have 0 or 1 element, because as far as I know, we do not run multiple Semi-supervised Engines in the same node and some nodes may not run the Semi-supervised Engine. I think this `agent_keys` vector might be a reflection of other considerations that I am not aware of. If this point was discussed in the meeting with Kone, could you kindly share the any insight regarding this please?

kimhanbeom: The `agent_key` vector size can be larger than 1. For example, in Tis project, we ran three Semi-supervised Engine on three nodes with the same customer id. Of course, cases like this project are not common, but I think it's good to have this configuration in mind.

CHANGELOG.md
kimhanbeom: done

src/graphql/customer.rs
sophie-cluml: Thank you for sharing.

sophie-cluml (approved):

src/graphql/node/crud.rs
```diff
@@ -192,24 +180,32 @@ async fn load(
...
-/// Returns the customer id of review node.
+/// Returns a customer id and the agent key(agent@hostname) list for the node corresponding to
+/// that customer id.
 ///
 /// # Errors
 ///
 /// Returns an error if the node profile could not be retrieved.
-#[allow(clippy::module_name_repetitions)]
-pub fn get_customer_id_of_node(db: &Store) -> Result<Option<u32>> {
+pub fn get_agent_key_by_customer_id(db: &Store) -> Result<HashMap<u32, Vec<String>>> {
```
sehkone, line 189: According to [Rust Naming Conventions](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter), the `get_` prefix is not recommended unless it's necessary. And, not a single key but multiple keys will be returned, so I think `agent_keys_by_customer_id` would be better.

kimhanbeom: done

src/graphql/customer.rs
```diff
@@ -9,11 +9,46 @@ use async_graphql::{
...
+    }
+
+    #[must_use]
+    pub fn target_agent_keys(&self) -> Vec<String> {
+        self.target_agent_keys.clone()
+    }
+
+    #[must_use]
+    pub fn networks(&self) -> database::HostNetworkGroup {
+        self.networks.clone()
```
sehkone, line 48: Is it impossible to return a reference, `&HostNetworkGroup`?

src/graphql/customer.rs
```diff
@@ -9,11 +9,46 @@ use async_graphql::{
...
+            networks,
+            target_agent_keys: agent_keys
+                .into_iter()
+                .filter(|s| s.starts_with(target_agent))
+                .collect(),
+        }
+    }
+
+    #[must_use]
+    pub fn target_agent_keys(&self) -> Vec<String> {
```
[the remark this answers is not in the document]
kimhanbeom: done

src/graphql/customer.rs
```diff
@@ -154,34 +191,36 @@ impl CustomerMutation {
...
-                    None
+
+        let mut changed_networks = Vec::new();
+        let store = crate::graphql::get_store(ctx).await?;
+        let mut map = store.customer_map();
+        map.update(i, &old, &new)?;
+
+        let mut hosts = vec![];
+        let mut networks = vec![];
+        let mut ip_ranges = vec![];
```
sehkone, line 202: These lines seem to be able to move into the below `if let Some(agent_keys)`.

sehkone (changes requested):

src/graphql/customer.rs
sehkone: Additionally, could you consider using the `fold` method instead of the `for` loop?

kimhanbeom: done

src/graphql/node.rs
```diff
@@ -12,13 +12,17 @@ use async_graphql::{
...
+pub use crud::agent_keys_by_customer_id;
 use database::Indexable;
 use input::NodeInput;
 use review_database as database;
 use roxy::Process as RoxyProcess;
 use serde::{Deserialize, Serialize};
 
+const SENSOR_AGENT: &str = "piglet";
+const UNSUPERVISED_AGENT: &str = "reconverge";
+pub(super) const SEMISUPERVISED_AGENT: &str = "hog";
```
sehkone, line 24: Since "semi-supervised" is an official term, we should use `SEMI_SUPERVISED`.

kimhanbeom: done

sehkone (approved):

Issue #393, opened 2025-01-21
Bump dependecies: axum to 0.8 and related to match axum 0.8


sophie-cluml:
As of now, axum 0.8 is released, but async-graphql-axum that supports axum 0.8 is not yet released. However, the new release of  async-graphql-axum will soon come out, as the support is merged to the main branch of the project https://github.com/async-graphql/async-graphql/issues/1652. When it is released, we can proceed to bump the dependencies.

PR #396, opened 2025-01-23
Update axum to 0.8

kimhanbeom:
async-graphql-axum is updated accordingly.

Close: #393

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,13 +8,11 @@ anyhow = "1"
 async-graphql = { version = "7", features = ["chrono", "string_number"] }
 async-graphql-axum = "7"
 async-trait = "0.1"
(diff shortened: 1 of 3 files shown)


sophie-cluml (approved):

sehkone:
@sophie-cluml @kimhanbeom,

Since axum-related crates are the ones we should keep an eye on, we need to specify their updates in commit messages.

kimhanbeom:
Added commit messages for version updates for axum and async-graphql.

sehkone:
> * Update axum to version 0.8.1.
> * Update async-graphql-axum to version 7.0.14.

Axum is updated to 0.8.x, not 0.8.1, which is currently specified in Cargo.toml. The commit message could be simplified to the following:

Update axum to 0.8

Its body can be like:

async-graphql-axum is updated accordingly.

Issue #273, opened 2024-08-21
Add documentation comments for GraphQL APIs, if they are missing


sophie-cluml:
It would be helpful to provide client-side friendly GraphQL API documentation, allowing client-side development to proceed without needing to understand the details of review-web. Developers could simply reference the GraphQL API schema, typically in SDL format. Some APIs have documentation, but not all are documented. So, I think this would mainly involve adding missing documentation comments in the relevant code.

sehkone:
Regarding the APIs to return values parsed from packets, we need to not only fill in the missing ones but also ensure their descriptions are legitimate according to the standard protocols. Since the current API names do not clearly indicate the meaning of the values they return, it would be better to elaborate on them with descriptive texts. I asked Parker and Ben to add proper descriptions to each API in Notion, and I think this issue should be addressed after that is completed.

sophie-cluml:
The content for the documentation comment will be drafted in the Notion page. @danbi2990, @pott-cho , and @kiwon94 are working on the draft.

danbi2990:
FWIW, the Notion document can be found by searching 'Attributes of Events'.

danbi2990:
I will create sub-issues for each event and proceed with `BlockListBootp` event myself. The remaining events will be assigned to @kiwon94 and @pott-cho, as discussed with @sophie-cluml.

Before proceeding, I'd like to confirm that the `duration` and `session_end_time` fields will remain unchanged, as discussed in discussion repository issue 328. I noticed several red comments on the Notion page suggesting changes to `last_time`, which appears to contradict the conclusion reached in that issue.

sehkone:
- Use `end_time` in sensors and stores
- Use either `end_time` or `duration` elsewhere, depending on the context

That is what we've concluded. Based on it, I think we have two things to do:

### Action Items

1. **Rename some remaining `last_time` to `end_time`**  
   - This hasn't been fully corrected yet.  
   - I’ve left a comment about this on the Notion page. (@syncpark, @kimhanbeom)

2. **Review usages of `duration` and `end_time`** (outside of sensors and stores, which already use only `end_time`)  
   - Evaluate case by case whether `duration` is appropriate.  
   - No need to rush—it's completely fine to take the time to do this properly.

sehkone:
This issue deals with GraphQL APIs, not syslog messages, so it's fine not to take those into account.

[...]

PR #408, opened 2025-03-04
[Draft] Add document for `BlockListBootp`

sophie-cluml:
Related: #273

This draft PR focuses on a single case: `BlockListBootp`, as a starting point for the #273. I think we can use this case as an  example for the remaining work.  

While working on this case, I noticed that several field names are missing both English names in the Notion document. I labeled those missing cases as TODO in the code. @sehkone Could you provide English names for these fields please, when you are available?


sophie-cluml:
Due to the https://github.com/aicers/review-web/issues/273#issuecomment-2716737689, this draft PR can be closed.

Issue #410, opened 2025-03-05
Fix unable to `updateNodeDraft` by "Missing configuration for agent key"


sophie-cluml:
One of the ways to reproduce the bug:

1. Add a node item with a unsupervised agent and a semi-supervised agent.
2. Apply the node.
3. Uncheck the unsupervised agent. Click "Save".
4. Edit the node to check a sensor agent. Click "Save".

The error message we get from the UI is below: 

> updateNodeDraft:2:5: Missing configuration for agent key: {sensor}

PR #412, opened 2025-03-06
Fix missing config error in `updateNodeDraft`

BLYKIM:
Previously, when an agent was newly added in `updateNodeDraft`, its configuration was expected to be present in `old_config_map`. To maintain consistency with `insertNode`, the logic was updated to return `None` instead of an error when a newly added agent does not exist in `old_config_map`.

Close: #410


sophie-cluml:
Let me close this PR, since the assignee of the issue changed.

Issue #405, opened 2025-02-19
Modify `removeTrustedDomain` GraphQL API to Support Multiple Removal Targets for Consistency


sophie-cluml:
By adjusting the current `removeTrustedDomain` GraphQL API to accept multiple removal targets, we can ensure consistency across GraphQL APIs that carry out removing actions.

PR #415, opened 2025-03-11
Support removing multiple trusted domains

sophie-cluml:
Close: #405

### DIFF
--- a/src/graphql/trusted_domain.rs
+++ b/src/graphql/trusted_domain.rs
@@ -86,20 +86,44 @@ impl TrustedDomainMutation {
         Ok(name)
     }
 
-    /// Removes a trusted domain, returning the old value if it existed.
-    #[graphql(
-        guard = "RoleGuard::new(Role::SystemAdministrator).or(RoleGuard::new(Role::SecurityAdministrator))"
-    )]
-    async fn remove_trusted_domain(&self, ctx: &Context<'_>, name: String) -> Result<String> {
-        {
-            let store = crate::graphql::get_store(ctx).await?;
-            let map = store.trusted_domain_map();
-            map.remove(&name)?;
+    /// Removes multiple trusted domains, returning the removed values.
+    ///
+    /// On error, some trusted domains may have been removed.
+    #[graphql(guard = "RoleGuard::new(Role::SystemAdministrator)
+        .or(RoleGuard::new(Role::SecurityAdministrator))")]
+    async fn remove_trusted_domains(
+        &self,
+        ctx: &Context<'_>,
+        #[graphql(validator(min_items = 1))] names: Vec<String>,
+    ) -> Result<Vec<String>> {
+        let store = crate::graphql::get_store(ctx).await?;
+        let map = store.trusted_domain_map();
+
+        let count = names.len();
+        let removed = names
+            .into_iter()
+            .try_fold(Vec::with_capacity(count), |mut removed, name| {
+                if map.remove(&name).is_ok() {
+                    removed.push(name);
+                    Ok(removed)
+                } else {
+                    Err(removed)
+                }
+            })
+            .unwrap_or_else(|r| r);
+
+        if removed.is_empty() {
+            return Err("None of the specified trusted domains was removed.".into());
         }
 
         let agent_manager = ctx.data::<Box<dyn AgentManager>>()?;
         agent_manager.broadcast_trusted_domains().await?;
-        Ok(name)
+
+        if removed.len() < count {
+            return Err("Some trusted domains were removed, but not all.".into());
+        }
+
+        Ok(removed)
     }
 }
 
@@ -173,21 +197,12 @@ mod tests {
             res.data.to_string(),
             r#"{insertTrustedDomain: "example2.org"}"#
         );
-
         let res = schema
-            .execute(r"{trustedDomainList{edges{node{name}}}}")
+            .execute(r#"mutation{insertTrustedDomain(name:"example3.org",remarks:"test")}"#)
             .await;
         assert_eq!(
             res.data.to_string(),
-            r#"{trustedDomainList: {edges: [{node: {name: "example1.com"}}, {node: {name: "example2.org"}}]}}"#
-        );
-
-        let res = schema
-            .execute(r#"mutation{removeTrustedDomain(name:"example1.com")}"#)
-            .await;
-        assert_eq!(
-            res.data.to_string(),
-            r#"{removeTrustedDomain: "example1.com"}"#
+            r#"{insertTrustedDomain: "example3.org"}"#
         );
 
         let res = schema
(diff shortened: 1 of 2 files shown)


sehkone (dismissed):

sehkone:
@sophie-cluml,

Could you rebase this?

sophie-cluml:
> @sophie-cluml,
> 
> Could you rebase this?

Thank you for notifying. Rebasing is done.

src/graphql/trusted_domain.rs
```diff
@@ -86,20 +86,27 @@ impl TrustedDomainMutation {
...
+        ctx: &Context<'_>,
+        names: Vec<String>,
+    ) -> Result<Vec<String>> {
+        let store = crate::graphql::get_store(ctx).await?;
+        let map = store.trusted_domain_map();
+
+        let removed: Vec<String> = names
+            .into_iter()
+            .filter_map(|name| map.remove(&name).ok().map(|()| name))
+            .collect();
```
sehkone, line 104: This doesn't properly respond when all items failed to be removed. In such cases, `removed` will be an empty vector and the API still returns `Ok`. Furthermore, I don't think it's ideal to ignore failures and just collect successfully removed ones. Most other APIs seem to handle exceptions that occur during the removal process. If there’s a specific rationale for this design, could you please share it with me?

sophie-cluml: Thank you for the feedback.

Unlike other removal endpoints, this particular endpoint calls `broadcast_trusted_domains`. So I initially considered the combination of database removal and broadcasting as a single atomic operation from the client's perspective. Based on that assumption, I thought it would be more consistent to not return an error during `map.remove()`, even if some items fail, in order to preserve the overall flow.

However, upon reflection, I realize this change has gone beyond the original scope of the issue. Since the existing implementation uses `map.remove(&name)?;`, it seems more appropriate to keep returning `Err` on failure during removal, as originally designed.

That said, I would appreciate some clarification on the following point:

> "I don't think it's ideal to ignore failures and just collect successfully removed ones. Most other APIs seem to handle exceptions that occur during the removal process."

From my understanding, our most common removal pattern looks like the following (sampled from `remove_allow_networks`):

```rust
let mut removed = Vec::<String>::with_capacity(ids.len());
for id in ids {
    let i = id.as_str().parse::<u32>().map_err(|_| "invalid ID")?;
    let key = map.remove(i)?;

    let name = match String::from_utf8(key) {
        Ok(key) => key,
        Err(e) => String::from_utf8_lossy(e.as_bytes()).into(),
    };
    removed.push(name);
}
```

To my understanding, this logic works as follows:

* If an error occurs during removal of the first item, **no items are removed**, and an `Err` is returned.
* If an error occurs midway, **items removed prior to the error remain removed**, and the function returns an `Err`.

If we apply the same logic to `remove_trusted_domains`, the expected behavior would be:

* If an error occurs on the first domain, nothing is removed, **a broadcast is not performed**, and an `Err` is returned.
* If an error occurs partway, domains before the error are removed, **a broadcast is performed** to align the semi-supervised engine with the DB state, and an `Err` is returned.

Could you let me know if this is the error handing behavior you expect for this endpoint as well please?

sehkone: > Unlike other removal endpoints, this particular endpoint calls `broadcast_trusted_domains`. So I initially considered the combination of database removal and broadcasting as a single atomic operation from the client's perspective. Based on that assumption, I thought it would be more consistent to not return an error during `map.remove()`, even if some items fail, in order to preserve the overall flow.
> 
> However, upon reflection, I realize this change has gone beyond the original scope of the issue. Since the existing implementation uses `map.remove(&name)?;`, it seems more appropriate to keep returning `Err` on failure during removal, as originally designed.

@sophie-cluml

This isn’t just a matter of scope—it’s about principle. Exceptions **must** be handled appropriately, and **must not** be **silently** ignored. Returning `Ok` when all removals fail is especially problematic, as it misleads the user with information that directly contradicts the actual outcome.

There are multiple valid ways to handle exceptions in such cases, and the choice depends on the context. For example:

- Stop at the first exception and return an `Err` (as `remove_trusted_domains` currently does).
- Iterate through all items, attempt to remove each one, and return an `Err` if any removal fails.
- Stop at the first failure, roll back any successful removals (i.e., restore previously removed items), and return an `Err`.
- Stop at the first failure and return an `Err` along with a list of items that failed to be removed.
- Iterate through all items, attempt to remove each one, and return an `Err` along with a list of items that failed to be removed.

The reason `broadcast_trusted_domains` takes its current approach is because it’s the simplest option that still ensures consistency. That approach can apply here as well.

sehkone: > Exceptions must be handled appropriately, and must not be silently ignored.

Of course, exceptions may be silently ignored in leaf-level code, such as in the UI.

sophie-cluml: Thank you for the clarification. I agree with "Exceptions must be handled appropriately, and must not be silently ignored." and https://github.com/aicers/review-web/pull/415#discussion_r2083491823.

Just to confirm, my question was:

> If we apply the same logic to `remove_trusted_domains`, the expected behavior would be:
>
> - If an error occurs on the first domain, nothing is removed, a broadcast is not performed, and an Err is returned.
> - If an error occurs partway, domains before the error are removed, a broadcast is performed to align the semi-supervised engine with the DB state, and an Err is returned.
>
> Could you let me know if this is the error-handling behavior you expect for this endpoint as well?

From your message:

> The reason `remove_allow_networks` takes its current approach is because it’s the simplest option that still ensures consistency. That approach can apply here as well.

I believe this is a "yes" to the question above. If this is not what you intended, I would appreciate your further clarification.

sehkone: > I believe this is a "yes" to the question above.

Correct

sophie-cluml: Thank you for guidance. The PR is adjusted.

src/graphql/trusted_domain.rs
```diff
@@ -86,20 +86,37 @@ impl TrustedDomainMutation {
...
+
+        for name in names {
+            if map.remove(name.as_str()).is_ok() {
+                removed.push(name);
+            } else {
+                if removed.is_empty() {
+                    return Err("None of the specified trusted domains was removed.".into());
+                }
+                let agent_manager = ctx.data::<Box<dyn AgentManager>>()?;
+                agent_manager.broadcast_trusted_domains().await?;
```
sehkone, line 112: These lines are unnecessarily repetitive and may distract the reader. I believe the following flow would be more intuitive and easier to follow:

1. Iterate over the names using `iter`.  
2. During the iteration, attempt to remove each item from the map. Stop at the first failure and return a `Result` containing the names that were successfully removed.  
3. Check the number of removed items. If there are any, call `broadcast_trusted_domains`.  
4. Finally, create and return another `Result` to end the method.

This structure helps readers understand the logic more clearly, without redundant code or fragmented control flow.  
To implement the above, I believe using `try_fold` is one of the good choices.

sophie-cluml: Thank you, done.

sehkone (approved):

Issue #414, opened 2025-03-10
Modify GraphQL APIs related to account


kimhanbeom:
Related issue: https://github.com/petabi/review-database/issues/408

With the addition of the `customer_ids` field to the `Account` structure in review-database,  The account-related GraphQL APIS for review-web should also be modified accordingly.

sophie-cluml:
I’d like to suggest a small alignment regarding the validation logic for `Account::customer_ids`.

Here’s how I currently view the intended behavior, from my point of view:

1. **SystemAdministrator** should always have access to *all customers*, which means their `customer_ids` should always be `None`. In other words, assigning specific customer ids to a SystemAdministrator should not be allowed.

2. For **SecurityAdministrator**, **SecurityManager**, and **SecurityMonitor**, `customer_ids` should not be `None`. They should always be `Some`, although an empty vector (i.e., `Some([])`) is acceptable.

I think the implementation in the linked PR mostly aligns with the logic above.

However, I noticed that for point 1, the validation seems to be implemented in `insert_account`, but not in `update_account`. For point 2, while validation exists in `insert_account`, it appears to be handled a bit differently in `update_account`. This could be due to a different interpretation of the required validation for `Account::customer_ids`.

I’d like to get us aligned on the expected validation rules. @kimhanbeom, could you share your perspective on this?

Also, since this involves customer-related logic, I believe @sehkone’s guidance could be especially helpful here. Would you mind reviewing and weighing in on the validation direction for `Account::customer_ids` please?

kimhanbeom:
If @sehkone has no further suggestions regarding the customer-related logic, it might be a good idea to proceed with @sophie's suggestion.

sehkone:
@sophie-cluml @kimhanbeom 

Sorry for my late reply. I don't quite agree with this point.

- System administrators don't always need access all customers. But they can if necessary.
- `Some([])` is meaningless from the user's perspective, whereas it should be acceptable on the server side.

sophie-cluml:
Understood.

PR #421, opened 2025-03-20
Modify GraphQL APIs related to account

kimhanbeom:
Close: #414

### DIFF
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -177,11 +177,15 @@ impl AccountMutation {
         theme: Option<String>,
         allow_access_from: Option<Vec<IpAddress>>,
         max_parallel_sessions: Option<u8>,
+        customer_ids: Option<Vec<u32>>,
     ) -> Result<String> {
         let store = crate::graphql::get_store(ctx).await?;
         let table = store.account_map();
         if table.contains(&username)? {
             return Err("account already exists".into());
+        }
+        if customer_ids.is_none() && role != Role::SystemAdministrator {
+            return Err("You are not allowed to access all customers.".into());
         }
         let allow_access_from = if let Some(ip_addrs) = allow_access_from {
             let ip_addrs = to_ip_addr(&ip_addrs);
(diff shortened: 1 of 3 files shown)


sophie-cluml (approved):

src/graphql/account.rs
```diff
@@ -177,12 +177,16 @@ impl AccountMutation {
...
         max_parallel_sessions: Option<u8>,
+        customer_ids: Option<Vec<u32>>,
     ) -> Result<String> {
         let store = crate::graphql::get_store(ctx).await?;
         let table = store.account_map();
         if table.contains(&username)? {
             return Err("account already exists".into());
         }
+        if customer_ids.is_none() && role != Role::SystemAdministrator {
+            return Err("You must provide at least one customer id to insert an account that is not in the System Administrator role.".into());
```
sehkone, line 188: @kimhanbeom @sophie-cluml 

We don't allow anyone other than system administrators to access all customers setting `customers_ids` to `None`. However, we don't say that users must provide at least one customer, because review-web allows an empty vector of customers. In that case, the user won't be able to access any customers, even though the UI should prevent this. Therefore, the message should be rephrased as:

You are not allowed to access all customers.

src/graphql/account.rs
```diff
@@ -284,10 +291,35 @@ impl AccountMutation {
...
+            let customer_ids_to_check = customer_ids
+                .as_ref()
+                .map_or(&account.customer_ids, |update_customer_ids| {
+                    &update_customer_ids.new
+                });
+
+            if customer_ids_to_check.is_none()
+                && role_to_check != database::Role::SystemAdministrator
+            {
+                return Err("You must provide at least one customer id to update an account that is not in the System Administrator role.".into());
```
sehkone, line 319: The same applies here.

kimhanbeom: done

src/graphql/account.rs
kimhanbeom: done

sehkone (approved):

Issue #439, opened 2025-04-28
Disallow the creation of a new system administrator


sophie-cluml:
From the comment https://github.com/aicers/review-web/issues/420#issuecomment-2735522639, we should disallow the creation of a new system administrator, meaning our system has only one system administrator.

sehkone:
@sophie-cluml and @pott-cho 

After implementing this feature in review-web, I think we should add an user-interactive feature to prevent doing such a thing.

Issue #420, opened 2025-03-18
Restrict SecurityAdministrator from Changing SystemAdministrator's Password in `update_account` Function


sophie-cluml:
### Background

A bug report from QA

### Detail

In the broader scope, a comprehensive review of role-based access control is planned, and related authorization code will be revised in the near future.  

However, I believe **SecurityAdministrator** can change the **SystemAdministrator**'s password in the `update_account` function is a clear security issue. So I think we can consider addressing this specific separately before the overall role-based access control review.

I think the following task is our minimum task:  

- Sub-task 1: Prevent **SecurityAdministrator** from changing the **SystemAdministrator**'s password.  

Additionally, allowing a **SecurityAdministrator** to modify the **SystemAdministrator**'s account information could be considered an inversion of authority. Therefore, I think the following task can also be included:  

- Sub-task 2: Prevent **SecurityAdministrator** from modifying **SystemAdministrator**'s account information.

sehkone:
Do you mean the following?

Prevent a security administrator from changing another security administrator’s password

sophie-cluml:
I meant preventing SecurityAdministrator from changing SystemAdministrator's password.

sehkone:
I suggest:

- Any system administrator can change any other user's information, including their password.
- All users except system administrators cannot change another user's information.

sehkone:
Additionally, I think it would be better to disallow the creation of a new system administrator.

sophie-cluml:
Hi octoaide, 

- I agree with https://github.com/aicers/review-web/issues/420#issuecomment-2735515202
- https://github.com/aicers/review-web/issues/420#issuecomment-2735522639 will be addressed in #439

PR #489, opened 2025-06-20
Disallow the creation of a new system administrator

pott-cho:
Close: #439 

* Prevent system administrator creation via `insert_account`
* Prevent system administrator update via `update_account`
  * not `SYSTEM_ADMINISTRATOR` -> `SYSTEM_ADMINISTRATOR` (x)
  * `SYSTEM_ADMINISTRATOR` -> not `SYSTEM_ADMINISTRATOR` (x) 
* Fix broken test cases due to function changes

### DIFF
--- a/src/graphql/account.rs
+++ b/src/graphql/account.rs
@@ -192,10 +192,6 @@ impl AccountMutation {
         max_parallel_sessions: Option<u8>,
         customer_ids: Option<Vec<ID>>,
     ) -> Result<String> {
-        if role == Role::SystemAdministrator {
-            return Err("Invalid role request".into());
-        }
-
         // Validate and normalize the username
         let normalized_username = validate_and_normalize_username(&username)
             .map_err(|e| format!("Invalid username: {e}"))?;
@@ -294,13 +290,6 @@ impl AccountMutation {
         for username in usernames {
             // Normalize the username for lookup (convert to lowercase)
             let normalized_username = username.to_lowercase();
-
-            if let Some(account) = map.get(&normalized_username)? {
-                if account.role == review_database::Role::SystemAdministrator {
-                    return Err("System administrator can not be removed".into());
-                }
-            }
-
             map.delete(&normalized_username)?;
             removed.push(normalized_username);
         }
@@ -1550,28 +1539,6 @@ mod tests {
 
     #[tokio::test]
     #[serial]
-    async fn remove_admin_account() {
-        // given
-        let original_review_admin = backup_and_set_review_admin();
-        assert_eq!(env::var(REVIEW_ADMIN), Ok("admin:admin".to_string()));
-        let schema = TestSchema::new().await;
-
-        // when
-        let res = schema
-            .execute(r#"mutation { removeAccounts(usernames: ["admin"]) }"#)
-            .await;
-
-        // then
-        assert_eq!(
-            res.errors.first().unwrap().message,
-            "System administrator can not be removed"
-        );
-
-        restore_review_admin(original_review_admin);
-    }
-
-    #[tokio::test]
-    #[serial]
     async fn default_account() {
         let original_review_admin = backup_and_set_review_admin();
         assert_eq!(env::var(REVIEW_ADMIN), Ok("admin:admin".to_string()));
@@ -1745,10 +1712,8 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn reset_admin_password_security_administrator() {
+    async fn reset_admin_password() {
         let schema = TestSchema::new().await;
-
-        // given
         let res = schema
             .execute(
                 r#"mutation {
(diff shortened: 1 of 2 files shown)


src/graphql/account.rs
```diff
@@ -168,6 +168,10 @@ impl AccountMutation {
         max_parallel_sessions: Option<u8>,
         customer_ids: Option<Vec<ID>>,
     ) -> Result<String> {
+        if role == Role::SystemAdministrator {
+            return Err("Invalid role request".into());
```
sophie-cluml, line 172: What do you think about something like 'not allowed' instead of 'invalid'? I think we are prohibiting creating a "SystemAdministrator" account by business policy. So I think it might be more of not allowing, than being invalid.

src/graphql/account.rs
```diff
@@ -266,6 +270,13 @@ impl AccountMutation {
         for username in usernames {
             // Normalize the username for lookup (convert to lowercase)
             let normalized_username = username.to_lowercase();
+
+            if let Some(account) = map.get(&normalized_username)? {
+                if account.role == review_database::Role::SystemAdministrator {
+                    return Err("System administrator can not be removed".into());
+                }
+            }
```
sophie-cluml, line 278: From my perspective, preventing the deletion of system administrator accounts seems to be a separate matter from disallowing the creation of new system administrator accounts, so these might be considered as distinct topics (maybe as a separate issue or GitHub discussion) ? For example, we might possibly consider whether to allow system administrator accounts to be removed without any restrictions, to enforce certain limits—such as only permitting removal if multiple system administrators exist—or to prevent any deletion of system administrator accounts altogether, as you suggested. I would appreciate your thoughts on this.

Additionally, I think `update_account` function could potentially create a new system administrator by updating the `role`. I believe this falls within the scope of issue #439, but I would like to hear your opinion on this as well.

pott-cho: Thank you for the thoughtful point. While creation and deletion can be seen as separate topics, I interpreted the comment in #439 — "our system has only one system administrator" — to mean that exactly one system administrator must always exist.
Since the administrator is injected externally at server startup, I believe removing both creation and deletion capabilities from the API is necessary to enforce this constraint.
With that in mind, I think handling deletion within the scope of issue #439 is appropriate.

I initially thought it belonged to issue #420, but handling it in issue #439 makes more sense.
I'll include it in the update.

sophie-cluml: I can see that you've created https://github.com/aicers/review-web/discussions/497. Thank you for that. Since this part is under discussion, could you revert code changes related to system administrator account deletion in this PR please?

src/graphql/account.rs
```diff
@@ -332,6 +343,13 @@ impl AccountMutation {
             }
         }
 
+        // Disallow update to SystemAdministrator
+        if role.as_ref().is_some_and(|role| {
+            (role.old == Role::SystemAdministrator) ^ (role.new == Role::SystemAdministrator)
+        }) {
+            return Err("Role not allowed.".into());
+        }
```
sophie-cluml, line 351: I think it would be more resource-efficient to validate the request parameters as early as possible, similar to what you did here: https://github.com/aicers/review-web/pull/489/files#diff-9ea84843fe763ff1a5f23c9b0823874914f17681621e588157dfcbe7ebb65982R171. Would you consider moving this validation logic closer to where the request parameters are checked, perhaps around L315 or L323?

sophie-cluml: Also, I think using ^ is a rather clever choice. However, I believe it would also prevent updating a system administrator user to a non-system administrator user. Since the deletion policy is still under discussion, I would like to suggest that we only disallow the creation of new system administrator accounts for now.

src/graphql/account.rs
pott-cho: Sure. I'll revert it.

src/graphql/account.rs
sophie-cluml: I think my above sentence was not clear enough. I thought it could be good to have something like:

```
        if role.as_ref().is_some_and(|role| {
            (role.old != Role::SystemAdministrator) && (role.new == Role::SystemAdministrator)
        }) {
            return Err("Role not allowed.".into());
        }
```

to disallow the creating of new system administrator accounts, but still allow deletion of system administrator accounts, via `updateAccount`?

pott-cho: Sure, I applied above the code.

sophie-cluml:
@pott-cho Could you take a look at https://github.com/aicers/review-web/actions/runs/15963557701/job/45019927304?pr=489#step:5:916 please?

pott-cho:
@sophie-cluml 
I haven’t yet been able to proceed with the implementation based on the direction discussed in the [discussion](https://github.com/aicers/review-web/discussions/497). 
While working on it, I found a few [suspicious parts](https://github.com/aicers/review-web/pull/478#discussion_r2181314206) in the existing code and was in the process of checking them.
I noticed that the PR has been approved—would it be okay to wrap it up in its current state?
If needed, I can clean up the commit messages and push an updated version.

pott-cho:
@sehkone 
Added the changes to the changelog.

sehkone:
@pott-cho, the commit messages you created don't follow our conventions. In such cases, combining them into a single commit would be appropriate. I would normally ask you to do so, but due to time constraints, I'll proceed to merge this by squashing the commits.

However, I hope you're well aware of this going forward.

Cc: @sophie-cluml

Issue #445, opened 2025-05-14
Improve Max Parallel Session Logic to Exclude Expired JWTs


sophie-cluml:
### Background

There is a max parallel session limit per account. When an abnormal logout occurs (e.g., browser refresh or closing the browser, causing the JWT to be lost from client browsers), the user is not able to log in again even after the JWT has expired. This is because the system currently checks only for the existence of the JWT in the access token map, regardless of expiration.

At the dev task meeting today, it was decided to address this issue with the following actions.

### Task

- During sign-in, validate the max parallel session limit based only on non-expired JWT tokens.
- After successful sign-in, remove any expired JWT tokens for the user from the access token map.

### Additional Note

I think both `sign_in` and `sign_in_with_new_password` functions should be reviewed and updated as needed.

sehkone:
> When an abnormal logout occurs (e.g., browser refresh or closing the browser, causing the JWT to be lost from client browsers), the user is not able to log in again even after the JWT has expired. This is because the system currently checks only for the existence of the JWT in the access token map, regardless of expiration.

@sophie-cluml, that's not correct.

The current code parses incoming tokens, verifies their signatures, decodes them into the `Claim` struct, and checks whether they are expired. Only after passing all of these steps does it check whether the token exists in the access token map.

As I mentioned in the meeting, the real issue is that the system currently counts expired tokens as well as valid ones when determining the number of concurrent sign-in sessions. That’s why the first task you mentioned is indeed necessary:

> - During sign-in, validate the max parallel session limit based only on non-expired JWT tokens.

The second task isn’t required to fix the issue itself, but it would be beneficial for better token maintenance and cleanup.

sophie-cluml:
It seems my English wording in the original message was unclear. Sorry for causing any misunderstanding.

Regarding the sentence,  

> This is because the system currently checks only for the existence of the JWT in the access token map, regardless of expiration.

What I intended to say is that the specific function `validate_max_parallel_sessions`, which is one of the validation functions called when user tries sign-in, currently checks only for the existence of the JWT in the access token map, regardless of expiration.

sehkone:
Thank you for the clarification. I'm glad to hear that we're already aligned.

PR #553, opened 2025-07-24
Improve max parallel session logic

kimhanbeom:
- Exclude expired JWTs from max parallel session checks in `signIn` and `signInWithNewPassword`.
- Automatically revoke expired tokens after successful sign-in using `revoke_expired_tokens`.
- Disable `Validation.validate_exp` in `decode_token` to allow decoding of all tokens, enabling expired token deletion.

Close: #445


sehkone:
@kimhanbeom, could you rebase this?

src/auth/jwt.rs
```diff
@@ -67,11 +67,9 @@ pub fn decode_token(token: &str) -> anyhow::Result<Claims> {
     let jwt_secret = JWT_SECRET
         .read()
         .map_err(|e| AuthError::ReadJwtSecret(e.to_string()))?;
-    let token_data = decode::<Claims>(
-        token,
-        &DecodingKey::from_secret(&jwt_secret),
-        &Validation::default(),
-    )?;
+    let mut validation = Validation::default();
+    validation.validate_exp = false; // Disable expiration validation
```
msk, line 71: No need to make `validation` mutable.

```rust
let validation = Validation {
    validate_exp: false,
    ..Validation::default()
};
```

src/graphql/account.rs
```diff
@@ -973,13 +976,15 @@ fn sign_in_actions(
...
     username: &str,
+    expiry_cutoff: i64,
 ) -> Result<AuthPayload> {
     let (token, expiration_time) =
         create_token(account.username.clone(), account.role.to_string())?;
     account.update_last_signin_time();
     account_map.put(account)?;
 
     insert_token(store, &token, username)?;
+    revoke_expired_tokens(store, expiry_cutoff, username)?;
```
msk, line 987: What do you think about moving this cleanup logic into a separate, scheduled background task instead of running it during the sign-in request? This would keep the sign-in endpoint consistently fast and prevent potential delays if there are many tokens in the database. It also simplifies error handling. A cleanup failure in a background job wouldn't result in an error in signing in.

src/auth/jwt.rs
kimhanbeom: Since `Validation::validate_signature` is scoped as `pub(crate)`, struct literal initialization with `..Validation::default()` isn’t supported. Therefore, I’ll keep the current implementation in the PR as-is.

src/graphql/account.rs
kimhanbeom: Processing this cleanup logic through a scheduled background task seems to be a more reasonable approach. Accordingly I’ll exclude this part from the current PR changes.

msk (approved):

src/graphql/account.rs
kimhanbeom: I created https://github.com/aicers/patio/issues/419

msk:
The change looks fine now.

Issue #547, opened 2025-07-22
Add logs for audit purposes


sae-bom:
Add logs for key user actions and system events needed for auditing.

PR #559, opened 2025-07-30
Add logs for audit purposes

sae-bom:
Closes: #547

Some cases from the initial draft on our private wiki are missing in the current code:
- Requests handled directly by the web UI without triggering any GraphQL calls
- Logs related to time series policies (could not locate the relevant code paths)
- Logs related to traffic filters (the name of the traffic filter could not be retrieved)
- Other functionalities that appear to be outside the scope of review-web

### DIFF
--- a/src/graphql/allow_network.rs
+++ b/src/graphql/allow_network.rs
@@ -6,12 +6,14 @@ use async_graphql::{
 use database::Direction;
 use review_database::{self as database, Store};
 use serde::{Deserialize, Serialize};
+use tracing::info;
 
 use super::{
     BoxedAgentManager, Role, RoleGuard,
     customer::{HostNetworkGroup, HostNetworkGroupInput},
 };
 use crate::graphql::query_with_constraints;
+use crate::info_with_username;
 
 #[derive(Default)]
 pub(super) struct AllowNetworkQuery;
@@ -64,11 +66,12 @@ impl AllowNetworkMutation {
             networks.try_into().map_err(|_| "invalid network")?;
         let value = review_database::AllowNetwork {
             id: u32::MAX,
-            name,
+            name: name.clone(),
             networks,
             description,
         };
         let id = map.put(value)?;
+        info_with_username!(ctx, "Allowlist {name} has been registered");
 
         apply_allow_networks(&store, ctx).await?;
         Ok(ID(id.to_string()))
@@ -99,6 +102,7 @@ impl AllowNetworkMutation {
                         Ok(key) => key,
                         Err(e) => String::from_utf8_lossy(e.as_bytes()).into(),
                     };
+                    info_with_username!(ctx, "Allowlist {name} has been deleted");
                     removed.push(name);
                     Ok(removed)
                 } else {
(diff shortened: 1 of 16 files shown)


sehkone:
@sae-bom,

Could you rebase this?

src/graphql/triage/policy.rs
```diff
@@ -119,6 +122,7 @@ impl TriagePolicyMutation {
         let store = crate::graphql::get_store(ctx).await?;
         let map = store.triage_policy_map();
         let id = map.put(triage)?;
+        info_with_username!(ctx, "Triage policy {name_clone} has been registered");
```
sehkone, line 125: When cloning a value is necessary, I think it’s better to preserve the original value for as long as possible, consuming only the cloned copies until the original must be used at the end. That’s what I meant in this case:

```rust
        let triage = database::TriagePolicy {
            id: u32::MAX,
            name: name.clone()
            ti_db,
            packet_attr: packet_attr_convert,
            confidence,
            response,
            creation_time: Utc::now(),
        };

        let store = crate::graphql::get_store(ctx).await?;
        let map = store.triage_policy_map();
        let id = map.put(triage)?;
        info_with_username!(ctx, "Triage policy {name} has been registered");
``` 

In this case, cloning is only needed once, so the two approaches don’t look very different. But in scenarios where multiple clones are required, this approach clearly improves readability for the reader.

sehkone (changes requested):

src/graphql/allow_network.rs
```diff
@@ -62,13 +64,15 @@ impl AllowNetworkMutation {
...
             networks.try_into().map_err(|_| "invalid network")?;
+        let name_clone = name.clone();
         let value = review_database::AllowNetwork {
             id: u32::MAX,
             name,
             networks,
             description,
         };
         let id = map.put(value)?;
+        info_with_username!(ctx, "Allowlist {name_clone} has been registered");
```
sehkone, line 75: Need the same approach I mentioned earlier

src/graphql/customer.rs
```diff
@@ -125,6 +129,7 @@ impl CustomerMutation {
             creation_time: Utc::now(),
         };
         let id = map.put(value)?;
+        info_with_username!(ctx, "Customer {name_clone} has been registered");
```
sehkone, line 132: The same case

src/graphql/network.rs
```diff
@@ -91,6 +96,7 @@ impl NetworkMutation {
             tag_ids,
         );
         let id = map.insert(entry)?;
+        info_with_username!(ctx, "Network {name_clone} has been registered");
```
sehkone, line 99: The same

src/graphql/triage/policy.rs
sae-bom: Thanks for the better approach! All done.

src/graphql/node/control.rs
```diff
@@ -24,6 +24,7 @@ impl NodeControlMutation {
         if !review_hostname.is_empty() && review_hostname == hostname {
             Err("cannot reboot. review reboot is not allowed".into())
```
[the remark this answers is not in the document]
sae-bom: I tried adding a few cases, but I’m still figuring out what should be considered meaningful, so I’m open to suggestions if you have a clearer idea.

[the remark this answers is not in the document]
sehkone: What you've added looks good to me. If anything is missing or incomplete, it can always be updated later when someone realizes it needs to be done. For now, this is good to go.

sehkone (approved):

sophie-cluml (approved):

Issue #494, opened 2025-06-23
Update review-database dependency to include RocksDB support for `csv_column_extra`


kimhanbeom:
Related issue: https://github.com/petabi/review-database/issues/349

The review-database crate is currently being updated to fetch `CsvColumnExtra` from RocksDB instead of PostgreSQL. This is part of the migration away from the PostgreSQL-based `csv_column_extra` table, which is scheduled for removal soon.

Although the update in review-database is still in progress, once it is completed, review-web should update its dependency to incorporate the change. Currently, review-web stores `CsvColumnExtra` in RocksDB using `add_csv_column_extra`, but still fetches it from PostgreSQL.

Steps to apply the change:
 - Update the review-database dependency to a commit that includes the changes from issue [#349](https://github.com/petabi/review-database/issues/349).
- Modify the code in review-web to use the updated method that fetches `CsvColumnExtra` from RocksDB instead of PostgreSQL.

msk:
This issue requires updating the review-database dependency to `rev = "8514f06"`.

sehkone:
@kimhanbeom, could you complete this issue?

PR #612, opened 2025-08-11
Migrate review-web to RocksDB-based Table methods

kimhanbeom:
Close: #494


kimhanbeom:
@minshao Could you please review this PR?

src/graphql/statistics.rs
minshao, line 1: `db.get_column_statistics` in `column_statistics` needs to be replaced with the one in RocksDB as well

minshao (approved):

PR #630, opened 2025-09-03
Replace `map_join!` macro with itertools' `join`

msk:
- Added `itertools` crate to `Cargo.toml`.
- Refactored `TriageResponse` logging to use `itertools::join` instead of the `map_join!` macro for improved readability and performance.

### DIFF
--- a/src/graphql/event/bootp.rs
+++ b/src/graphql/event/bootp.rs
@@ -1,5 +1,6 @@
 use async_graphql::{Context, Object, Result, StringNumber};
 use chrono::{DateTime, Utc};
+use itertools::Itertools;
 use review_database::event as database;
 
 use super::{ThreatLevel, TriageScore, country_code, find_ip_customer, find_ip_network};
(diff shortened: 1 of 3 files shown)


sehkone:
The reason I introduced the `map_join!` macro is that it serves better performance and readability in cases like `graphql::event::dhcp::BlocklistDhcp::class_id`. The current implementation looks like this:

```rust
self.inner
    .class_id
    .iter()
    .map(|x| format!("{x:02x}"))
    .collect::<Vec<String>>()
    .join(":")
```

Using `map_join!`, the same logic can be simplified to:

```rust
map_join!(self.inner.class_id.iter(), ":", |x| "{x:02x}")
```

There are several similar cases scattered throughout the codebase. I tried `itertools`, but couldn’t find a clean way to cover these cases. My plan is to create another issue to apply this macro consistently across those places.  

Could you share your thoughts on this approach?

sehkone:
> I added another commit to handle the cases you mentioned as well as other string joins using itertools.

Oh! This looks so nice. I appreciate what you've done.

sehkone (approved):

sehkone:
@sophie-cluml and @sae-bom,

Let’s make sure to keep this in mind for future work.

Issue #729, opened 2025-12-05
Enhancements to the Triage Policy


kimhanbeom:
## Background 
The structure of the Triage Policy in review-database has been updated as follows:

1. Add a `customer_id: Option<u32>` field to `TriagePolicy`
2. Rename the existing `TriageExclusionReason` enum to `ExclusionReason`
3. create a new `TriageExclusionReason` struct that includes `id`, `name`, `description`, and an embedded `ExclusionReason`. 
    - Store this new struct in th new rocksdb map `triage_exclusion_map`
4. Changed the `TriagePolicy` field `ti_db: Vec<TriageExclusionReason>` to `triage_exclusion_id: Vec<u32>`.
    - In other words, it now references only the IDs of `TriageExclusionReason` stored in `triage_exclusion_map`.
5. Renamed fields previously using `ti_db` in Triage-related structs and functions to either `triage_exclusion_id` or `triage_exclusion`

Reference: https://github.com/aicers/review-database/pull/636

After these changes, the `TriagePolicy` structures in review-database are as follows:
```rust
  #[derive(Clone, Deserialize, Serialize)]
  pub struct TriagePolicy {
      pub id: u32,
      pub name: String,
      pub triage_exclusion_id: Vec<u32>, 
      pub packet_attr: Vec<PacketAttr>,
      pub confidence: Vec<Confidence>,
      pub response: Vec<Response>,
      pub creation_time: DateTime<Utc>,
      pub customer_id: Option<u32> 
  }
  
  #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
  pub struct TriageExclusionReason {
      pub id: u32,
      pub name: String,
      pub exclusion_reason: ExclusionReason,
      pub description: String,
  }
  
  #[derive(Clone, PartialEq, Deserialize, Serialize)]
  pub enum ExclusionReason {
      IpAddress(HostNetworkGroup),
      Domain(Vec<String>),
      Hostname(Vec<String>),
      Uri(Vec<String>),
  }
  ```
Based on these changes in review-database, the same modifications must also be applied to review-web as outlined below.

## Task
1. Implement new GraphQL APIs for the newly added `TriageExclusionReason` struct to support insert/remove/update/list operations.
   - This struct is stored in the `triage_exclusion_map` RocksDB table, so reference it during implementation.
2. Update existing Triage-related GraphQL APIs to align with the modified structures:
   - `TriagePolicyList`
      - Add a new parameter `customer_id: Option<ID>`.
[...]

PR #742, opened 2025-12-18
Enhance triage policy and exclusion GraphQL

kimhanbeom:
Close: #729

### DIFF
--- a/src/graphql/triage/policy.rs
+++ b/src/graphql/triage/policy.rs
@@ -271,7 +271,7 @@ impl TriagePolicyMutation {
         Ok(ID(id.to_string()))
     }
 
-    /// Removes triage policies, returning the IDs that no longer exist.
+    /// Removes triage policies, returning the names that no longer exist.
     ///
     /// On error, some triage policies may have been removed.
     #[graphql(guard = "RoleGuard::new(Role::SystemAdministrator)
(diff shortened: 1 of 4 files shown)


sophie-cluml (approved):

src/graphql/triage/policy.rs
```diff
@@ -71,36 +91,129 @@ async fn load(
...
     async fn insert_triage_policy(
         &self,
         ctx: &Context<'_>,
         name: String,
-        ti_db: Vec<TriageExclusionReasonInput>,
+        triage_exclusion_id: Vec<ID>,
         packet_attr: Vec<PacketAttrInput>,
         confidence: Vec<ConfidenceInput>,
         response: Vec<ResponseInput>,
+        customer_id: Option<ID>,
```
sehkone: It looks like that the method validates `customer_id` can be successfully parsed but doesn't check whether it exists. Is it good?

src/graphql/triage/policy.rs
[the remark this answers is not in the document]
kimhanbeom: I’ll update the comment to reflect that the method returns names.

kimhanbeom:
I also updated it to check triage policy references when a customer is deleted.

src/graphql/triage/policy.rs
sehkone: I don't understand. This API is one of the GraphQL APIs accessible to any users. How can we make it sure that the input is always legitimate?

Issue #775, opened 2026-01-30
Add Validation to Prevent Duplicate Network Ranges per Customer


pott-cho:
### Backgrounds
When adding a network under [Admin > Customer], the system currently allows registering the same IP even if the network names differ. This behavior should be restricted so that the same customer cannot register the same IP more than once.

### Required Change
Whenever a network is added to a customer, the system must check whether the same value already exists in the incoming data.
If a duplicate value is found, the network creation should be rejected.

pott-cho:
I confirmed that the requirements of this issue are already implemented in the code. Even if duplicate values are included in newly registered customer network inputs, they are silently removed internally without raising an error.

**types.rs**
Sorting and deduplication are performed in `HostNetworkGroup::new()`.
During this process, `hosts`, `networks`, and `ip_ranges` are processed with `sort_unstable()` followed by `dedup()`, so duplicate values are removed without any error.

**customer.rs**
`HostNetworkGroup::new()` is called during the input conversion step.
In `TryFrom<&HostNetworkGroupInput>`, the values are parsed and then passed as `Ok(Self::new(hosts, networks, ip_ranges))`, meaning duplicates are removed before being stored in the DB.

Therefore, in this issue, we will only add tests to make the current behavior explicit and ensure it is properly covered.

PR #789, opened 2026-02-19
Add testcode to network group behavior

pott-cho:
Closes: #775

I added test code to clarify the behavior of customer network add/update.
The tests in this PR clarify the following behaviors:

* When adding customer networks, the `networkGroup` fields (`hosts`, `networks`, `ranges`) are stored in sorted order.
* When adding customer networks, duplicate values within a single `networkGroup` are removed without an error.
* When adding customer networks, identical `networkGroup` values across different network entries are allowed (for both insert and update).

### DIFF
--- a/src/graphql/customer.rs
+++ b/src/graphql/customer.rs
@@ -1074,4 +1074,264 @@ mod tests {
                 .contains("still referenced by block network")
         );
     }
+
+    #[tokio::test]
+    async fn ensure_network_group_items_are_sorted() {
+        // The values inside networkGroup are stored in deterministic sorted order (per type-defined ordering), regardless of input order.
+        let schema = TestSchema::new().await;
+
+        let res = schema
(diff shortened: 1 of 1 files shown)


src/graphql/customer.rs
```diff
@@ -1074,4 +1074,232 @@ mod tests {
...
+
+        let actual = res
+            .data
+            .into_json()
+            .expect("response data should be valid JSON");
+        assert_eq!(actual, expected);
+    }
+
+    #[tokio::test]
+    async fn allow_duplicate_network_groups_on_insert() {
```
sophie-cluml: I think adding assertions that verify the network entries are actually preserved as intended (e.g., both n1 and n2 exist) would improve the test’s purpose. Similar assertions might be helpful to  `allow_duplicate_network_groups_on_update` as well.

PR #847, opened 2026-05-09
Add eventListWithTriage resolver (#842)

sehkone:
## Summary

Adds a new `eventListWithTriage` GraphQL query that returns events passing the standard filter regardless of policy match, with optional inline triage scoring and exclusions. The resolver implements the four-step structure from the issue: standard filter pass, inline exclusion pre-cut, attach-only policy scoring, and cursor pagination with `first + 1` / `last + 1` accounting performed on surviving events.

The query supports two caller patterns with an identical contract:

- Interactive / corpus B trigger: callers populate `triage` with `policies` and/or `exclusions`; events that score nothing keep `triageScores: null` and remain in the connection.
- Cadence corpus A fill: callers pass `triage = null`; the resolver returns every event passing the standard filter and the cadence runner applies its own exclusion logic before INSERT.

New GraphQL input types — `EventStandardFilterInput`, `EventTriageInput`, `EventTriagePolicyInput`, `EventTriageExclusionInput` — are introduced alongside the resolver. `EventStandardFilterInput` intentionally omits `triagePolicies` so the inline `triage.policies` argument is the only way to express scoring intent. `EventTriageExclusionInput` is flattened per populated field; an empty exclusion object is rejected with a GraphQL error during resolver-side input validation. The existing `eventList`, `eventTriageList`, and `EventListFilterInput` are left untouched (add-only).

`Confidence.threatCategory` and `ConfidenceInput.threatCategory` are also made nullable to track the underlying `review-database` change.

Role guard matches `eventList` / `eventTriageList`: callable by `SystemAdministrator | SecurityAdministrator | SecurityManager | SecurityMonitor`.

Closes #842

## Test plan

- [x] `triage = null` returns the same edges, cursors, and `hasNextPage` / `hasPreviousPage` as `eventList` for an equivalent filter and pagination args
- [x] An event that no policy scores above its `minimum_score` is still returned with `triageScores: null` (not dropped, not an empty list)
- [x] `triage = { exclusions: [...], policies: null }` removes the matched events from both the connection edges and `totalCount`
- [x] When excluded events outnumber surviving ones in a raw page, the resolver keeps scanning until `first + 1` / `last + 1` survivors are collected or the time range is exhausted; no empty page is returned while later survivors remain
- [x] An exhausted range with no survivors yields empty edges with `hasNextPage: false`
- [x] A query attempting `eventListWithTriage(filter: { triagePolicies: [...] }, ...)` fails at GraphQL parse/validation time
- [x] An `EventTriageExclusionInput` with multiple populated fields produces the same cut-event set as the equivalent multi-element list of single-field objects
- [x] An empty `EventTriageExclusionInput` (all fields omitted, or supplied as null / empty list / empty group) returns a GraphQL error
- [x] `totalCount` equals the edge sum after fully draining the connection (standard-filter-passing minus inline-excluded)
- [x] `cargo clippy --no-default-features --features auth-jwt --all-targets -- -D warnings` passes
- [x] `cargo clippy --no-default-features --features auth-mtls --all-targets -- -D warnings` passes
- [x] `cargo fmt -- --check --config group_imports=StdExternalCrate` passes
- [x] `cargo test --no-default-features --features auth-jwt` passes
- [x] `cargo test --no-default-features --features auth-mtls` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,32 @@ This file documents recent notable changes to this project. The format of this
 file is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and
 this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added `eventListWithTriage` GraphQL query that returns events passing the
+  standard filter regardless of policy match, with optional inline triage
+  scoring and exclusions. The new resolver supports two caller patterns:
+  the interactive/corpus B trigger path that supplies the `triage` argument
+  with inline `policies` and/or `exclusions`, and the cadence corpus A fill
+  path that calls with `triage = null`. Events that score nothing keep
+  `triageScores: null` and remain in the connection. Inline exclusions cut
+  events from the connection and from `totalCount`. Cursor pagination
+  matches `eventList` semantics, and `first + 1` / `last + 1` accounting is
+  performed on surviving (post-exclusion) events. Accepts the new input
+  types `EventStandardFilterInput`, `EventTriageInput`,
+  `EventTriagePolicyInput`, and `EventTriageExclusionInput`. Callable by
+  `SystemAdministrator | SecurityAdministrator | SecurityManager |
+  SecurityMonitor`.
+
+### Changed
+
+- `Confidence.threatCategory` and `ConfidenceInput.threatCategory` are now
+  nullable (`ThreatCategory` -> `Option<ThreatCategory>`) following the
+  underlying `review-database` change that made `Confidence::threat_category`
+  optional.
+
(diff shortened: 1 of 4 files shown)


sehkone:
**[Reviewer Round 2]**

The Round 1 item is resolved. I verified that `convert_event_triage_exclusions` now treats `ipAddress` as populated only when at least one of `hosts`, `networks`, or `ranges` is non-empty, and the regression assertion covers `ipAddress: { hosts: [], networks: [], ranges: [] }`.

I re-reviewed the updated diff against #842. The resolver keeps the standard filter path on `triage_policies: None`, applies inline exclusions before scoring, uses `score_against_policies` only as an attach step, and paginates/counts surviving events. I do not see any remaining blockers from this round.

sehkone:
[Review Verdict Round 2: APPROVED]

